PR #29, opened 2024-11-29
Support for docker login and registry mirror

kvinwang:
This PR adds support for config docker registry mirror and login credentials.

The registry mirror is configured in the `teepod.toml`:
```toml
[cvm]
docker_registry = "http://10.0.2.2:5000/"
```

The login credentials can be configured as optional fields when deplying the app:
image
Note
- KMS is required to use this feature.
- The TOKEN value must be put in the encrypted env and fill the env name in the `Token Key` field.


tolak:
@kvinwang Can explain more why KMS is needed here?

kvinwang:
> Can you explain more why KMS is needed here?

The token need to be encrypted in front end and decryped in CVM. So it require the KMS to get the app pubkey before deployment.

PR #33, opened 2024-12-03
Added GetInfo & ResizeVm APIs to enhance management capabilities

Leechael:


teepod/src/main_service.rs
```diff
@@ -241,6 +241,48 @@ impl TeepodRpc for RpcHandler {
...
+            .manifest()
+            .context("failed to read manifest")?;
+        if let Some(vcpu) = request.vcpu {
+            manifest.vcpu = vcpu;
+        }
+        if let Some(memory) = request.memory {
+            manifest.memory = memory;
+        }
+        if let Some(disk_size) = request.disk_size {
+            manifest.disk_size = disk_size;
```
kvinwang, line 293: Note: Resizing the disk size will not take effect

teepod/src/app.rs
```diff
@@ -302,4 +302,51 @@ impl App {
...
+            }),
+            app_url: info.instance_id.as_ref().map(|id| {
+                format!(
+                    "https://{id}-{}.{}:{}",
+                    gw.tappd_port, gw.base_domain, gw.port
+                )
+            }),
+            app_id: info.manifest.app_id,
+            instance_id: info.instance_id,
+        })
```
kvinwang: Looks like this should be extracted to a function that can be reused in `list_vms` and `get_vm`.

teepod/src/app.rs
```diff
@@ -302,4 +302,51 @@ impl App {
             })
             .collect())
     }
+
+    pub fn get_vm(&self, id: &str) -> Option<pb::VmInfo> {
+        let state = self.state.lock().unwrap();
+        let vm = state.monitor.iter_vms().find(|vm| vm.info().manifest.id == id)?;
```
kvinwang: The internal structure is a map, so linear searching is not required.

teepod/src/main_service.rs
Leechael: Let me add a comment regarding this situation and it's ok we only updating the manifest here for now.

kvinwang (approved):
LGTM

PR #40, opened 2024-12-05
Teepod: assign cid when starting cvm and recycle it on stopped

Leechael:


kvinwang:
Is there any problem that this PR intends to resolve?

Leechael:
> Is there any problem that this PR intends to resolve?

Closed. It's a concept not yet ready.

PR #46, opened 2024-12-10
GetMeta APIs

Leechael:
This PR have added the GetMeta API for Teepod, TProxy, and KMS. This API retrieves the configured behavior for each core service:
- **Teepod**: Returns the associated TProxy and KMS addresses along with the maximum capacity.
- **TProxy**: Currently returns statistics only, as non-critical configuration options should not be exposed.
- **KMS**: Detects whether `allow_any_upgrade` is enabled to determine if an associated CVM can process upgrades.

These APIs allow the creation of a standalone service that connects to multiple DStack deployments and activates various features based on metadata.


kvinwang (approved):
LGTM

PR #47, opened 2024-12-11
Add Go SDK implementation

Hyodar:
Hi there!

We're now using Dstack in some of our TEE experiments in Nethermind - more recently https://github.com/NethermindEth/teeception!
The codebase is in Go, so we implemented a quick Go Dstack SDK based on the existing TS and Python implementations. If that's interesting to you, happy to contribute here. Feel free to let me know if there's anything else needed.

This PR adds a Go SDK implementation effectively equivalent to the TS and Python ones, as well as a simple README and unit tests.

Using it is mostly similar to the other ones as well - here's an example:
```go
package main

import (
	"context"
	"fmt"
	"log/slog"

	"github.com/Dstack-TEE/dstack/sdk/go/tappd"
)

func main() {
	client := tappd.NewTappdClient(
		// tappd.WithEndpoint("http://localhost"),
		// tappd.WithLogger(slog.Default()),
	)

	deriveKeyResp, err := client.DeriveKey(context.Background(), "/", "test", nil)
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(deriveKeyResp) // &{-----BEGIN PRIVATE KEY--- ...

	tdxQuoteResp, err := client.TdxQuote(context.Background(), []byte("test"), tappd.SHA256)
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(tdxQuoteResp) // &{0x0000000000000000000 ...

	rtmrs, err := tdxQuoteResp.ReplayRTMRs()
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(rtmrs) // map[0:00000000000000000 ...
}
```
```

### CODE REVIEW

File: sdk/go/tappd/client.go [hunk 1]
```diff
@@ -0,0 +1,274 @@
...
+	}
+	return &response, nil
+}
+
+// Sends a TDX quote request to the Tappd service.
+func (c *TappdClient) TdxQuote(ctx context.Context, reportData []byte, hashAlgorithm QuoteHashAlgorithm) (*TdxQuoteResponse, error) {
+	hexData := hex.EncodeToString(reportData)
+	if hashAlgorithm == RAW {
+		if len(hexData) > 128 {
+			return nil, fmt.Errorf("report data is too large, it should be less than 128 characters when hash_algorithm is raw")
```
kvinwang:
```suggestion
			return nil, fmt.Errorf("report data is too large, it should be at most 64 bytes when hashAlgorithm is RAW")
```

kvinwang (commented):
Thank you for the awesome implementation. Looks excellent overall, with one minor note. @Leechael, please review.

sdk/go/tappd/client.go
```diff
@@ -0,0 +1,274 @@
...
+
+	var response DeriveKeyResponse
+	if err := json.Unmarshal(data, &response); err != nil {
+		return nil, err
+	}
+	return &response, nil
+}
+
+// Sends a TDX quote request to the Tappd service.
+func (c *TappdClient) TdxQuote(ctx context.Context, reportData []byte, hashAlgorithm QuoteHashAlgorithm) (*TdxQuoteResponse, error) {
```
Leechael: Can we add `TdxQuoteWithHashAlgorithm`? The idea is to keep the API design simpler. We would have:

`TdxQuote("test")` - Uses the default hash algorithm (sha512)
`TdxQuoteWithHashAlgorithm("test", tappd.RAW)` - Allows specifying a custom hash algorithm

Golang doesn't support optional parameters, so this would be a better approach.

sdk/go/tappd/client.go
```diff
@@ -0,0 +1,274 @@
...
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	return io.ReadAll(resp.Body)
+}
+
+// Derives a key from the Tappd service. If altNames is empty or nill, it will
+// not be included inthe request.
+func (c *TappdClient) DeriveKey(ctx context.Context, path string, subject string, altNames []string) (*DeriveKeyResponse, error) {
```
Leechael: Can we add more for it?
- `DeriveKeyWithSubject(path, subject)`
- `DeriveKeyWithSubjectAndAltNames(path, subject, altNames)`

Go doesn't support optional parameters, so this would be a better approach.

sdk/go/tappd/client.go
Leechael: @kvinwang It looks like he's copying from the original implementation (refs [1](https://github.com/Dstack-TEE/dstack/blob/master/sdk/js/src/index.ts#L232), [2](https://github.com/Dstack-TEE/dstack/blob/master/sdk/python/src/dstack_sdk/tappd_client.py#L137)). Should we consistently specify the data size before or after hex encoding?

kvinwang: > @kvinwang It looks like he's copying from the original implementation (refs [1](https://github.com/Dstack-TEE/dstack/blob/master/sdk/js/src/index.ts#L232), [2](https://github.com/Dstack-TEE/dstack/blob/master/sdk/python/src/dstack_sdk/tappd_client.py#L137)). Should we consistently specify the data size before or after hex encoding?

I pointed it out because the argument the caller passed in was bytes, so the error reported to the caller should be a hint to the caller about what's wrong with the argument.

Hyodar: I agree! This is from the other implementations, but since in this case we're enforcing a bytes representation, it makes more sense to refer to the bytes length. As far as I understand there's no need to standardize errors here

sdk/go/tappd/client.go
Hyodar: How do you feel about an options struct? Maybe this can make things less confusing vs. separate methods. For example:
```go
// parameters can be omitted here
client.DeriveKey(ctx, tappd.DeriveKeyOptions{
    Path: "",
    Subject: "",
    AltNames: []string{},
})
```
We could also have a default `DeriveKey` and this could be `DeriveKeyWithOptions`.

sdk/go/tappd/client.go
Hyodar: Agreed as well! In this case it seems very natural to have both methods. Do you foresee any additional optional parameters to this method? If yes, maybe a similar solution to `DeriveKey` could be applied as well, but I think it'd be overkill in this case.

sdk/go/tappd/client.go
Hyodar: Actually, now that I see the Python implementation again, I do think it'd be better to be consistent on the bytes data size - if the user passes a string, this is not a hex string but a string to be UTF-8 encoded. It may be more understandable to the user to relate his string to the byte size than the hex encoding length.

sdk/go/tappd/client.go
Leechael: The API allows all parameters to be empty; however, it is recommended to include at least one parameter, specifically `Path`, as a unique identifier or salt. I believe we can consistently guide users to utilize the SDK with best practices by default.

sdk/go/tappd/client.go
Hyodar: Done in [864e4c2](https://github.com/Dstack-TEE/dstack/pull/47/commits/864e4c207907036232c760fce84c89cfbc45b696)

sdk/go/tappd/client.go
Hyodar: Alright, that's understandable. Done in [44106a9](https://github.com/Dstack-TEE/dstack/pull/47/commits/44106a9f4e62e92b4a086066dd541c7de57c8563)

sdk/go/tappd/client_test.go
```diff
@@ -124,7 +124,7 @@ func TestTdxQuoteRawHash(t *testing.T) {
 
 	// Test too large raw hash
 	largeData := make([]byte, 65)
-	_, err = client.TdxQuote(context.Background(), largeData, tappd.RAW)
+	_, err = client.TdxQuote(context.Background(), largeData)
```
Leechael: The change is unnecessary because the length check applies only to `RAW`.

Hyodar: Makes total sense, ended up skipping this - tests failed anyway. Fixed in [9eaf1dc](https://github.com/Dstack-TEE/dstack/pull/47/commits/9eaf1dc20619561541f580b56fedfb574f8d345d)

PR #91, opened 2025-01-10
KMS as Tapp

kvinwang:
This PR implements an onchain governed KMS which supports deploying itself as a Tapp on dstack. See the [doc](https://github.com/Dstack-TEE/dstack/blob/kms-onchain/kms/README.md) for the details.

Key rotation is not implemented in this PR. Since dstack support base image upgrading, I think we can implement key rotation in future versions.

@Leechael Please pay attention to the RPC changes in KMS and Tappd.

The reproducibility of the Tapp docker image is not done yet. Here is a temporary alternative which builds the image on-site:
```yaml
services:
  # Helios is a Ethereum light client
  helios:
    build:
      context: .
      dockerfile_inline: |
        FROM rust:1.80-alpine@sha256:1f5aff501e02c1384ec61bb47f89e3eebf60e287e6ed5d1c598077afc82e83d5
        WORKDIR /app
        RUN apk add --no-cache git build-base openssl-dev protobuf protobuf-dev perl
        RUN git clone https://github.com/a16z/helios && \
            cd helios && \
            git checkout 5c61864a167c16141a9a12b976c0e9398b332f07
        RUN cd helios && cargo build --release
        WORKDIR /app/helios
    command: [
      "./target/release/helios",
      "ethereum",
      "--network", "mainnet",
      "--checkpoint", "0xbee4f32f91e62060d2aa41c652f6c69431829cfb09b02ea3cad92f65bd15dcce",
      "--rpc-bind-ip", "0.0.0.0",
      "--rpc-port", "8545",
      "--consensus-rpc", "https://ethereum.operationsolarstorm.org",
      "--execution-rpc", "https://ethereum-rpc.publicnode.com"
    ]

  # Auth API is a webhook server that authenticates KMS instances and Apps launches
  auth-api:
    build:
      context: .
      dockerfile_inline: |
        FROM node:18-alpine@sha256:06f7bbbcec00dd10c21a3a0962609600159601b5004d84aff142977b449168e9
        WORKDIR /app

        RUN apk add --no-cache git
        RUN git clone https://github.com/Dstack-TEE/dstack.git && \
            cd dstack && \
            git checkout 4d9011d027af993e17a3fb8182a7249409b78a2e
        WORKDIR /app/dstack/kms/auth-eth
        RUN npm install && \
            npx hardhat typechain && \
            npx tsc --project tsconfig.json
        CMD node dist/src/main.js
    environment:
      - HOST=0.0.0.0
      - PORT=8000
      - ETH_RPC_URL=http://helios:8545
      - KMS_CONTRACT_ADDR=0x515cd0dd984Adc75EE8625131FcC3d2D91f7F140
    depends_on:
      - helios

  # KMS handles the TEE Remote Attestation
  kms:
    build:
      context: .
      dockerfile_inline: |
        FROM rust:1.80-alpine@sha256:1f5aff501e02c1384ec61bb47f89e3eebf60e287e6ed5d1c598077afc82e83d5
        WORKDIR /app
        RUN apk add --no-cache git build-base openssl-dev protobuf protobuf-dev
        RUN git clone https://github.com/Dstack-TEE/dstack.git && \
            cd dstack && \
            git checkout 4d9011d027af993e17a3fb8182a7249409b78a2e
        RUN cd dstack/kms && cargo build --release -p kms
        WORKDIR /app/dstack
        CMD ["./target/release/kms"]
    volumes:
      - kms-volume:/etc/kms
      - /var/run/tappd.sock:/var/run/tappd.sock
    ports:
      - 8000:8000
    depends_on:
      - auth-api

volumes:
  kms-volume:

```


dstack-types/src/lib.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+pub struct LocalConfig {
+    pub kms_url: Option<String>,
+    pub tproxy_url: Option<String>,
+    pub pccs_url: Option<String>,
+    pub docker_registry: Option<String>,
+    pub host_api_url: String,
+}
```
MoeMahhouk, line 90: for what purpose and where is this struct being used?

key-provider-client/src/host.rs
```diff
@@ -0,0 +1,53 @@
...
+    let response_length = u32::from_be_bytes(response_length);
+    let mut response = vec![0; response_length as usize];
+    tcp_stream
+        .read_exact(&mut response)
+        .await
+        .context("Failed to read response")?;
+    let response: QuoteResponse =
+        serde_json::from_slice(&response).context("Failed to deserialize response")?;
+    Ok(response)
+}
```
MoeMahhouk, line 53: if this is the client sending to gramine-sealing-key-provider directly, then we should probably also adjust it with the prefixed size exchange too. Because currently, it doesn't expect sending or receiving the size of the quote/payload.
Or do you have like a server in between that mediates between the two and handles the size prefixed payload exchange?

kms/auth-eth/contracts/KmsAuth.sol
```diff
@@ -0,0 +1,215 @@
...
+            return (false, "App not registered");
+        }
+
+        // Check aggregated MR and image measurements
+        if (
+            !allowedAggregatedMrs[bootInfo.mrAggregated] &&
+            !allowedImages[bootInfo.mrImage]
+        ) {
+            return (false, "Neither aggregated MR nor image is allowed");
+        }
```
MoeMahhouk, line 198: Maybe having those condition handled separately would be more explicit

key-provider-client/src/host.rs
kvinwang: > if this is the client sending to gramine-sealing-key-provider directly

Yes

> Because currently, it doesn't expect sending or receiving the size of the quote/payload.

[It does](https://github.com/MoeMahhouk/gramine-sealing-key-provider/blob/6b4658d3d2158615b6c78879ad61612885372712/src/server.rs#L52-L58)

This PR is tested running without problem with the key-provider built [here](https://github.com/Dstack-TEE/dstack/blob/kms-onchain/key-provider-build/Dockerfile)

dstack-types/src/lib.rs
kvinwang: The information is generated in host teepod and passed into CVM via a `config.json`.
Then it is used in CVM:
   - `kms_url` is used to connect to KMS to get app keys.
   - `pccs_url` is used during verifying the quote of key-provider.
   - `tproxy_url` is used to setup the tproxy net if enabled.
   - `docker_registry` is used as the official docker hub registry cache.
   - `host_api_url` is a VSOCK address used to notify teepod of the CVM boot progress.

kms/auth-eth/contracts/KmsAuth.sol
kvinwang: The logic is: if either condition is met, it is permitted. So it might can not address separately?

MoeMahhouk: ah, I see. I focused more on the error message and thought if it should fail if only one of them is missing.
But if that is the logic, then it is all good

key-provider-client/src/host.rs
MoeMahhouk: ah yes, my bad. got it confused with something else. 
Then all good.

dstack-types/src/lib.rs
MoeMahhouk: I have follow up questions here:
- does it make sense to have an attestation with this initial configuration hash passed by the host such that the dev can make sure the initial boot was with the correct config? Because IIUC, this config is not part of the CVM's measurements (RTMR0-2). So probably using either extended within RTMR[3] or user-report-data upon requesting attestation quote to verify the boot-up process of the CVM, correct?
- if the answer to the above question is yes, then could you refer me to that piece please?
- slightly unrelated question just for my understanding: if a TApp is deployed on a similar CVM base image and same config. How does the dev differentiate their instance from others trying to forge the same identity? Does this require pre-deployment setup such as calculating the TApp measurements and deploying AppAuth for each TApp. Then the light client within the CVM image would do checking the owner and compose hash of the App and then pulls it from the docker_registry? Also, vice-versa, does the CVM needs to verify the identity of the owner prior the light client checking and pulling the TApp container? I am asking this question because not sure if the dev/owner is doing the deployment themselves on bare-metal or there is an intermediary host, like Azure/GCP, that initiate the deployments and has control over the initial host provided config.

kvinwang: Very good questions.   
Answers to question 1 and 2:  
We currently don't measure the LocalConfig. Instead of trusting these URLs, we do more concrete verifications individually.  
- For `kms_url`, instead of trusting the URL, the CVM connects to KMS by RA-TLS. The KMS verifies CVM's RA quote. And the CVM measures the root public key of KMS to RTMR3, and the public key is also registered on-chain. Users can compare the root public key in RTMR3 against the on-chain value to ensure the CVM gets keys from the given KMS.  
- For `pccs_url`, the URL is always not trusted, but verified by the root CA certificate in the verification library.  
- For `tproxy_url`, it is also protected by bidirectional RA-TLS verification.  
- For `docker_registry`, the URL is also not trustable. The best practice is to always use image hashes in the `docker-compose.yaml` so that the docker engine would guarantee the image integrity after pulling the images.  
- For `host_api_url`, we don't need to trust it as it is used to emit boot progress and act as a proxy for requests to the key-provider.

Regarding question 3:  
Before deploying a Tapp CVM, an `app-id` is required to be registered in the KmsAuth contract and assign with an AppAuth contract to control which compose hash is allowed for the Tapp.  
CVMs request app keys from KMS at boot time. The KMS asks the KmsAuth/AppAuth whether the compose-hash and other MRs are allowed to boot. If either the KmsAuth or AppAuth rejects the boot request, the KMS will not distribute the keys and the CVM will abort the boot.

kms/auth-eth/src/ethereum.ts
```diff
@@ -0,0 +1,54 @@
...
+    hex = hex.startsWith('0x') ? hex.slice(2) : hex;
+
+    // Pad hex string to 64 characters (32 bytes)
+    hex = hex.padStart(sz * 2, '0');
+
+    // Add '0x' prefix back
+    return '0x' + hex;
+  }
+
+  async checkBoot(bootInfo: BootInfo, isKms: boolean): Promise<BootResponse> {
```
kvinwang, line 30: Review hint: This is where the KMS ask the smart contract for permission.

kms/src/onboard_service.rs
```diff
@@ -0,0 +1,269 @@
...
+    fn construct(context: CallContext<'_, OnboardState>) -> Result<Self> {
+        Ok(OnboardHandler {
+            state: context.state.clone(),
+        })
+    }
+}
+
+impl OnboardRpc for OnboardHandler {
+    async fn bootstrap(self, request: BootstrapRequest) -> Result<BootstrapResponse> {
+        let keys = Keys::generate(&request.domain).context("Failed to generate keys")?;
```
kvinwang, line 47: Review hint: Generating KMS root keys

kms/src/main_service.rs
```diff
@@ -195,9 +220,61 @@ impl KmsRpc for RpcHandler {
...
-            ca_cert: self.state.inner.root_ca.cert.pem(),
-            allow_any_upgrade: self.state.inner.config.allow_any_upgrade,
+            ca_cert: self.state.inner.root_ca.pem_cert.clone(),
+            allow_any_upgrade: self.state.inner.config.auth_api.is_dev(),
+            k256_pubkey: self.state.inner.k256_key.to_bytes().to_vec(),
+            bootstrap_info,
+        })
+    }
+
+    async fn get_kms_key(self) -> Result<KmsKeyResponse> {
```
kvinwang, line 230: Review Hint: Here the running KMS shares the keys to new KMS instance.

kms/src/main_service/upgrade_authority.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+impl AuthApi {
+    pub async fn is_app_allowed(&self, boot_info: &BootInfo, is_kms: bool) -> Result<BootResponse> {
+        match self {
+            AuthApi::Dev { dev } => Ok(BootResponse {
+                is_allowed: true,
+                reason: "".to_string(),
+                tproxy_app_id: dev.tproxy_app_id.clone(),
+            }),
```
kvinwang, line 53: Review Hint: Allow any app boot for local dev instance.

kms/src/main_service/upgrade_authority.rs
```diff
@@ -0,0 +1,79 @@
...
+                } else {
+                    "bootAuth/app"
+                };
+                let url = url_join(&webhook.url, path);
+                let response = client.post(&url).json(&boot_info).send().await?;
+                if !response.status().is_success() {
+                    bail!("Failed to check boot auth: {}", response.text().await?);
+                }
+                Ok(response.json().await?)
+            }
```
kvinwang, line 67: Review Hint: Asking the webhook for boot authority.

kms/src/onboard_service.rs
```diff
@@ -0,0 +1,269 @@
...
+}
+
+async fn quote_keys(p256_pubkey: &[u8], k256_pubkey: &[u8]) -> Result<(Vec<u8>, Vec<u8>)> {
+    let p256_hex = hex::encode(p256_pubkey);
+    let k256_hex = hex::encode(k256_pubkey);
+    let content_to_quote = format!("dstack-kms-genereted-keys-v1:{p256_hex};{k256_hex};");
+    let hash = keccak256(content_to_quote.as_bytes());
+    let report_data = pad64(hash);
+    let res = tapp_quote(report_data).await?;
+    Ok((res.quote, res.event_log.into()))
```
kvinwang, line 230: Review Hint: Generating quote for the root keys. The quote would be published onchain.

ra-tls/src/attestation.rs
```diff
@@ -179,6 +144,50 @@ impl Attestation {
...
+            sha256(&[&key_provider_info])
+        };
+        let mr_aggregated = sha256(&[
+            &td_report.mr_td,
+            &rtmrs[0],
+            &rtmrs[1],
+            &rtmrs[2],
+            &mr_key_provider,
+        ]);
+        let mr_image = sha256(&[&td_report.mr_td, &rtmrs[1], &rtmrs[2]]);
```
kvinwang, line 172: Review Hint: Calculating the `mr_aggregated` and `mr_image`.

ra-tls/src/attestation.rs
```diff
@@ -193,39 +202,198 @@ impl Attestation {
...
+                anyhow::bail!("RTMR mismatch");
+            }
         }
-        Ok(())
+        Ok(VerifiedAttestation {
+            quote: self.quote,
+            raw_event_log: self.raw_event_log,
+            event_log: self.event_log,
+            report,
+        })
```
kvinwang, line 307: Review Hint: This is only place creating `VerifiedAttestation` instance. So when you see `VerifiedAttestation` in the codebase, it must be verified.

tdxctl/src/fde_setup.rs
```diff
@@ -181,44 +184,111 @@ impl SetupFdeArgs {
...
+            .await
+            .context("Failed to get app key")?;
+        {
+            let (_, ca_pem) = x509_parser::pem::parse_x509_pem(tmp_ca.ca_cert.as_bytes())
+                .context("Failed to parse ca cert")?;
+            let x509 = ca_pem.parse_x509().context("Failed to parse ca cert")?;
+            let id = hex::encode(x509.public_key().raw);
+            let provider_info = KeyProviderInfo::new("kms".into(), id);
+            emit_key_provider_info(&provider_info)?;
+        };
```
kvinwang, line 243: Review Hint (**Important**): Requesting app keys from kms here. The pubkey in the ca cert, used by remote endpoint of the TLS, is written to RTMR3 as provider info.

tdxctl/src/fde_setup.rs
```diff
@@ -181,44 +184,111 @@ impl SetupFdeArgs {
...
+            .tls_client_key(cert_pair.key_pem)
+            .tls_ca_cert(tmp_ca.ca_cert.clone())
+            .cert_validator(Box::new(|cert| {
+                let Some(cert) = cert else {
+                    bail!("Missing server cert");
+                };
+                let Some(usage) = cert.special_usage else {
+                    bail!("Missing server cert usage");
+                };
+                if usage != "kms:rpc" {
```
kvinwang, line 220: Review Hint: A real kms only issues certs with `usage=kms:rpc` to themselves to server rpc, nerver issues it to apps.

tdxctl/src/fde_setup.rs
```diff
@@ -181,44 +184,111 @@ impl SetupFdeArgs {
...
+            .context("Failed to get sealing key")?;
+        // write to fs
+        let app_keys =
+            gen_app_keys_from_seed(&provision.sk).context("Failed to generate app keys")?;
+        let keys_json = serde_json::to_string(&app_keys).context("Failed to serialize app keys")?;
+        fs::write(self.app_keys_file(), keys_json).context("Failed to write app keys")?;
+
+        // write to RTMR
+        let provider_info = KeyProviderInfo::new("local-sgx".into(), hex::encode(provision.mr));
+        emit_key_provider_info(&provider_info)?;
```
kvinwang, line 272: Review Hint: For local key provider, the sgx `mr_enclave` is written to RTMR3. The sgx quote is verified in fn `get_sealing_key`.

teepod/src/host_api_service.rs
```diff
@@ -40,4 +41,19 @@ impl HostApiRpc for HostApiHandler {
...
+            bail!("Key provider is not enabled");
+        }
+        let response = get_key(request.quote, key_provider.address, key_provider.port)
+            .await
+            .context("Failed to get sealing key from key provider")?;
+
+        Ok(GetSealingKeyResponse {
+            encrypted_key: response.encrypted_key,
+            provider_quote: response.provider_quote,
+        })
```
kvinwang, line 57: Review Hint: Request from one CVM, delegate to the sgx key provider.

tdxctl/src/tboot.rs
```diff
@@ -88,12 +119,41 @@ impl<'a> Setup<'a> {
...
+            .maybe_pccs_url(self.local_config.pccs_url.clone())
+            .tls_client_cert(client_cert)
+            .tls_client_key(client_key.serialize_pem())
+            .tls_ca_cert(ca_cert)
+            .tls_built_in_root_certs(false)
+            .tls_no_check(self.app_keys.tproxy_app_id == "any")
+            .cert_validator(Box::new(move |cert| ra_validator.validate(cert)))
+            .build()
+            .into_client()
+            .context("Failed to create RA client")?;
```
kvinwang, line 156: Review Hint: Connect to tproxy here. If the KMS(smart contract) allow to connect to any tproxy, then we don't validate remote cert, otherwise, do bidirectional ra verification, and ensure the remote app-id matches the one KMS returned.

kvinwang:
@h4x3rotab I've added some review hints.

kms/auth-eth/contracts/AppAuth.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: MIT
```
kvinwang, line 1: Review Hint: And please pay attention to the smart contracts.

kms/auth-eth/contracts/KmsAuth.sol
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    // Function to deregister a KMS device ID
+    function deregisterKmsDeviceId(bytes32 deviceId) external onlyOwner {
+        allowedKmsDeviceIds[deviceId] = false;
+        emit KmsDeviceIdDeregistered(deviceId);
+    }
+
+    // Function to check if KMS is allowed to boot
+    function isKmsAllowed(
```
Ruteri, line 162: Not something for right now, but something we have spent a lot of time discussing is censorship resistance of de-registering measurements.
The worst case for kms is one where:
* a vulnerability is discovered, and the corresponding measurement de-registered
* however, existing kms instances are blocked from processing new blocks, and therefore don't actually see the de-registration transaction
* an attacker can spend as long as they need in order to extract the root secret, since the kms instance will never see the de-registration transaction
* the de-registration might as well not have happened at all

To counteract this scenario, the kms instance needs to make sure it's not being censored — neither in the short or in the long term. What we have arrived at as a possible solution is interacting onchain (kms at startup and then periodically requiring a random nonce to appear onchain), and limiting the number of actions allowed without forcing an onchain interaction. This way you would never allow an unlimited attempts at breaking a kms instance, since there's only so many attempts before you need to either reboot the instance (which will require onchain interaction) or are blocked by the action limit (which also requires onchain interaction)

kms/src/crypto.rs
```diff
@@ -0,0 +1,25 @@
+use anyhow::{Context, Result};
+use k256::ecdsa::{RecoveryId, Signature, SigningKey};
+use sha3::{Digest, Keccak256};
+
+use ra_tls::kdf;
+
+pub(crate) fn derive_k256_key(
```
Ruteri, line 7: I think it would be worth commenting on how apps are prevented from accessing each others keys

kms/src/main.rs
```diff
@@ -28,6 +35,41 @@ struct Args {
...
+    async fn index() -> RawHtml<&'static str> {
+        RawHtml(include_str!("www/onboard.html"))
+    }
+    #[rocket::get("/finish")]
+    fn finish(shutdown: Shutdown) -> &'static str {
+        shutdown.notify();
+        "OK"
+    }
+
+    if !kms_config.onboard.auto_bootstrap_domain.is_empty() {
```
Ruteri, line 51: I wonder if each instance should be whitelisted on chain before being allowed to onboard
The added visibility might be well worth it right?

ra-rpc/src/client.rs
```diff
@@ -1,50 +1,138 @@
...
+    tls_client_cert: Option<String>,
+    tls_client_key: Option<String>,
+    tls_ca_cert: Option<String>,
+    #[builder(default = true)]
+    tls_built_in_root_certs: bool,
+    pccs_url: Option<String>,
+    cert_validator: Option<CertValidator>,
+}
+
+impl RaClientConfig {
```
Ruteri, line 41: While I don't think using ratls is an issue, we have lately been moving away from relying on RA. Rather we exchange the regular TLS certificates along with their attestations (an example alternative implementation would have each kms instance register their locally generated TLS cert+attestation on boot). If registration prior to onboarding is anyway something you'd want, then this could simplify the implementation a bit

kms/auth-eth/contracts/KmsAuth.sol
kvinwang: That's an excellent consideration.

Let me see if I understand the proposed solution correctly:
The KMS node needs to interact with the chain both at startup and periodically by sending transactions, to verify it has access to up-to-date chain data.

I'm wondering if we could also consider using trusted time sources like NTS as an additional verification mechanism? Specifically, we could compare the block timestamp with the trusted timestamp - if the difference is within an acceptable threshold, it would help ensure the node is operating with fresh data rather than being fed stale blocks by an attacker.

kms/src/crypto.rs
kvinwang: The app-id is simply placed into the derivation path. I've moved the logic to this function for clarity.
https://github.com/Dstack-TEE/dstack/pull/91/commits/01b814e0725e28509a4d9eec621f11f90829a94f

kms/src/main.rs
kvinwang: The device-id (which is implemented as a hash of the ppid) is whitelisted. Do you think this is enough?

ra-rpc/src/client.rs
kvinwang: We need the RA either at the TLS level, HTTP level, or the Application level.
Each has its advantages and disadvantages.
- The benefits of the Application level are that it's easy to implement in different languages and web frameworks.
- The benefits of the TLS level are that it is easier to abstract, the attestation stuff is not needed to defined in each RPC definition schema; once a TLS connection is attested, all transferred data over the connection is reliable.
  We can easily reuse the same framework across several services, so far:
  - KMS: onboarding RPC, key providing RPC
  - tproxy: registering RPC
  - CVM: Inter-App Secure Communication (WIP)

The client side can simply use a `curl` command with an RA-TLS enabled cert to request those different services.

We may do the abstraction at the Application level, but that requires considerations about potential MITM attacks which are still complex.

kms/auth-eth/contracts/KmsAuth.sol
Ruteri: >  The KMS node needs to interact with the chain both at startup and periodically by sending transactions, to verify it has access to up-to-date chain data.

I think that's one way of doing it, yes

> I'm wondering if we could also consider using trusted time sources like NTS

I think so. Interacting with the blockchain is of course involved and costly — it'd be best to avoid it. With NTS specifically it would be interesting, because we are avoiding relying on local time — only considering what we are getting from NTS and the timestamp of the current chain head. I guess the thing to make sure is that NTS can't be replayed/delayed in a way that defeats the verification (but my guess is that it shouldn't be — as longs as it's https it could only be delayed, but never reordered) and the verification is as good as NTS itself (pretty good if you ask me in this scenario).

For what it's worth I think resolving this issue could be done at a later time
Cc @MoeMahhouk since you were thinking about nts recently

kms/src/main.rs
Ruteri: I'd guess that onboarding to kms (sharing the secret) is the most security-critical component, so I'd think that it'd be worth having an onchain trace of it happening — ideally sharing the secret requires an interactive onchain challenge (which would also force up-to-date view of the chain), but I'd imagine at least that all honest KMS instances would make it known that a given device requested the kms key (be it on chain or elsewhere) for transparency

ra-rpc/src/client.rs
Ruteri: One nice thing about doing it at the Application level is that it's more front-and-center, while ratls does hide the details a bit

> CVM: Inter-App Secure Communication (WIP)

Do keep in mind establishing the ratls channel does take considerably longer (~1s for Azure for example, if you need to fetch pccs it's also going to take a while). Performance considerations is one thing that pushed us towards exchanging attested tls certs beforehand, and using regular tls for all but the initial connection. Ratls is I think fine for registering & other non-performance critical use cases, but inter-app communication could need the connections to be established faster

kms/auth-eth/contracts/KmsAuth.sol
kvinwang: We are going to merge this initial version. Future improvements will be tracked in issue: https://github.com/Dstack-TEE/dstack/issues/115

kms/src/main.rs
kvinwang: We are going to merge this initial version. Future improvements will be tracked in issue: https://github.com/Dstack-TEE/dstack/issues/114

ra-rpc/src/client.rs
kvinwang: We are going to merge this initial version. Future improvements will be tracked in issue: https://github.com/Dstack-TEE/dstack/issues/113

PR #98, opened 2025-01-21
make docker restart always and run as daemon

0xshawn:
Updates

1. Make docker always restart in docker compose file
2. Make the local provider run as daemon by add `-d` argument to docker compose up
3. Add docker log options to limit the log file size


kvinwang (approved):
LGTM

PR #123, opened 2025-03-03
docs: add tproxy setup guide & faq

Leechael:


docs/faq.md
```diff
@@ -0,0 +1,26 @@
+# FAQ
+
+## CVM status turns to `exited` immediately
+
+You can see an error message in CVM's stderr output:
```
kvinwang: Need to append `&ch=stderr` to the log URL.

docs/tproxy.md
```diff
@@ -0,0 +1,49 @@
...
+
+You need to get a Cloudflare API Key and ensure the API can manage this domain.
+
+You can check your Cloudflare API key and get `cf_zone_id` using this command:
+
+```shell
+curl -X GET "https://api.cloudflare.com/client/v4/zones" -H "Authorization: Bearer <API_KEY>" -H "Content-Type: application/json" | jq .
+```
+
+Open your `certbot.toml`, and update these fields:
```
kvinwang, line 21: The certbot is now a built-in feature in tproxy. It can be configured in `tproxy.toml`

Leechael: Is the standalone Certbot still available? Is the initialization and renewal process fully automatic?

kvinwang: > Is the standalone Certbot still available?

Yes.

> Is the initialization and renewal process fully automatic?

When using the tporxy built-in certbot, it will exit after the cert is renewed. We can combine with external auto-restart, such as docker's `restart=always` to make it fully automatic.

The standalone certbot wont notify the tproxy when cert get renewed.

Leechael: > it will exit after the cert is renewed

And it will cause existing TCP connections to drop?

kvinwang: > And it will cause existing TCP connections to drop?

Yes, approximately once per quarter.

Leechael: As I checked, the certbot became a built-in feature after this PR: https://github.com/Dstack-TEE/dstack/pull/107. Since it's a big change, the documentation needs to be completely rewritten.

I think this is too much for this PR and it should be a separate task. This PR is proposed as a quick fix for the missing part of the doc, based on the tested behavior.

kvinwang: Make sense

kvinwang (approved):

PR #140, opened 2025-04-01
tappd: Don't reuse connections in healthy check client

kvinwang:
This fixes the error in the CVM console log:
```
tappd[643]: 2025-03-19T07:31:23.259508Z ERROR tappd: Health check request failed
```


h4x3rotab (approved):
lgtm

PR #144, opened 2025-04-04
API renaming for Client SDK

kvinwang:
Continuing PR #142: Client SDK renaming for easier review.

The old version of go sdk is not removed because it might used by downstream using repo url.
Js and python sdk are updated in place.


sdk/js/src/index.ts
```diff
@@ -222,24 +228,60 @@ export function send_rpc_request<T = any>(endpoint: string, path: string, payloa
...
+export class DstackClient {
   private endpoint: string
+  private logger: Console
 
-  constructor(endpoint: string = '/var/run/tappd.sock') {
+  constructor(endpoint: string = '/var/run/dstack.sock') {
+    this.logger = console
     if (process.env.DSTACK_SIMULATOR_ENDPOINT) {
-      console.debug(`Using simulator endpoint: ${process.env.DSTACK_SIMULATOR_ENDPOINT}`)
+      this.logger.debug(`Using simulator endpoint: ${process.env.DSTACK_SIMULATOR_ENDPOINT}`)
```
Leechael: Better switch to `INFO` because when working on customer support, we need the line to check if they built the image incorrectly.

Leechael (approved):
lgtm

sdk/js/src/index.ts
kvinwang: Done

PR #156, opened 2025-04-23
fix deploy scripts

Leechael:


gateway/dstack-app/docker-compose.yaml
```diff
@@ -27,7 +27,6 @@ services:
       - CF_API_TOKEN=${CF_API_TOKEN}
       - CF_ZONE_ID=${CF_ZONE_ID}
       - MY_URL=${MY_URL}
-      - BOOTNODE_URL=${BOOTNODE_URL}
```
kvinwang: Better keep it there, because we may want to deploy multiple nodes in production.

gateway/dstack-app/entrypoint.sh
```diff
@@ -86,7 +86,7 @@ rpc_domain = "gateway.$SRV_DOMAIN"
 run_in_dstack = true
 
 [core.sync]
-enabled = true
+enabled = false
```
kvinwang: Let's allow multiple nodes mode.

kvinwang (approved):
Great!

PR #161, opened 2025-05-01
Feat: Add rust crate for dstack client

nlok5923:
This pr add rust crate for dstack client. It contains all the relevant apis required for building an app on dstack.


sdk/rust/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "dstack-rust"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+ethers = "2.0"
```
tuddman: ethers was deprecated over a year ago: https://github.com/gakonst/ethers-rs/issues/2667
```suggestion
alloy = { version = "0.15", features = ["signers", "signer-local"] }
```

sdk/rust/src/ethereum.rs
```diff
@@ -0,0 +1,10 @@
+use crate::dstack_client::GetKeyResponse;
+use ethers::signers::LocalWallet;
```
tuddman: `LocalWallet` has changed to https://docs.rs/alloy/latest/alloy/signers/local/index.html

nlok5923: Fixed

sdk/rust/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+hex = "0.4"
+http = "0.2"
+http-client-unix-domain-socket = "0.1.1"
+reqwest = { version = "0.11", features = ["json"] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+sha2 = "0.10.9"
+
+[dev-dependencies]
+evidence_api = { git = "https://github.com/cc-api/evidence-api.git", branch = "main" }
```
Leechael: Suggest to change: use [dcap-qvl](https://github.com/Phala-Network/dcap-qvl) instead. We opened an issue (https://github.com/cc-api/evidence-api/issues/143) and submitted a PR (https://github.com/cc-api/evidence-api/pull/144) to evidence-api, but it seems no one is maintaining it anymore.

sdk/rust/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "dstack-rust"
+version = "0.1.0"
+edition = "2024"
```
Leechael: please add the authors

sdk/rust/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "dstack-rust"
```
kvinwang: I favor retaining `dstack-sdk` as the name to ensure consistency across js and python implementations.

sdk/rust/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+http = "0.2"
+http-client-unix-domain-socket = "0.1.1"
+reqwest = { version = "0.11", features = ["json"] }
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
+sha2 = "0.10.9"
+
+[dev-dependencies]
+dcap-qvl = "0.1.0"
+tokio = { version = "1", features = ["full"] }
```
kvinwang: As dstack is primarily a Rust project, it would be beneficial to add this SDK to the workspace as a member, so that we can centrally manage dependencies. So, let's delare the dependencies in the workspace Cargo.toml and reference then here for example:
sha2.workspace = true

sdk/rust/README.md
```diff
@@ -0,0 +1,112 @@
...
+
+```toml
+[dependencies]
+dstack-rust = { git = "https://github.com/Dstack-TEE/dstack.git", package = "dstack-rust" }
+```
+
+## Basic Usage
+
+```rust
+use dstack_client::DstackClient;
```
kvinwang: Do you mean `use dstack_rust::DstackClient;` ?

sdk/rust/src/dstack_client.rs
```diff
@@ -0,0 +1,270 @@
...
+        };
+
+        DstackClient {
+            base_url,
+            endpoint,
+            client,
+        }
+    }
+
+    async fn send_rpc_request(&self, path: &str, payload: &Value) -> Result<Value, Box<dyn Error>> {
```
kvinwang: Is it possible to use generic instead of dynamic typing?
Like:
```rust
async fn send_rpc_request<S: Serialize, D: DeserializeOwned>(&self, path: &str, payload: &S) -> anyhow::Result<D>
```

sdk/rust/src/dstack_client.rs
```diff
@@ -0,0 +1,270 @@
...
+                Ok(res.1)
+            }
+        }
+    }
+
+    pub async fn get_key(
+        &self,
+        path: Option<String>,
+        purpose: Option<String>,
+    ) -> Result<GetKeyResponse, Box<dyn Error>> {
```
kvinwang: Using `anyhow::Error` should be better than `Box<dyn Error>`.

sdk/rust/src/dstack_client.rs
```diff
@@ -0,0 +1,270 @@
...
+        Ok(())
+    }
+
+    pub async fn get_tls_key(
+        &self,
+        subject: Option<String>,
+        alt_names: Option<Vec<String>>,
+        usage_ra_tls: bool,
+        usage_server_auth: bool,
+        usage_client_auth: bool,
```
kvinwang: Let's implement a builder pattern for the arguments to prevent potential breaking changes and enhance code readability of the caller site, as we anticipate expanding support for additional fields in this API later.

For example, the crate `bon` is used in other crates in this project.
```
#[bon::Builder]
struct TlsKeyConfig {
...
}
pub async fn get_tls_key(&self, config: TlsKeyConfig)
```

sdk/rust/Cargo.toml
nlok5923: Added

nlok5923:
> @nlok5923 please consider accepting [RizeLabs#3](https://github.com/RizeLabs/dstack/pull/3) that addresses most of @kvinwang 's comments so this can PR can be merged upstream.

Thanks for the pr @tuddman merged it

sdk/rust/src/dstack_client.rs
```diff
@@ -0,0 +1,272 @@
...
+        });
+        let response = self.send_rpc_request("/GetKey", &data).await?;
+        let response = serde_json::from_value::<GetKeyResponse>(response)?;
+
+        Ok(response)
+    }
+
+    pub async fn get_quote(&self, report_data: Vec<u8>) -> Result<GetQuoteResponse> {
+        if report_data.is_empty() || report_data.len() > 64 {
+            return Err(anyhow!("Invalid report data length"));
```
kvinwang:
```suggestion
            anyhow::bail!("Invalid report data length");
```

sdk/rust/src/dstack_client.rs
```diff
@@ -0,0 +1,272 @@
...
+                let json_payload = serde_json::to_value(payload)?;
+                let res = unix_client
+                    .send_request_json::<Value, Value, Value>(
+                        path,
+                        Method::POST,
+                        &[("Content-Type", "application/json")],
+                        Some(&json_payload),
+                    )
+                    .await?;
+                Ok(serde_json::from_value(res.1)?)
```
kvinwang:
```suggestion
                let res = unix_client
                    .send_request_json::<_, _, Value>(
                        path,
                        Method::POST,
                        &[("Content-Type", "application/json")],
                        Some(&payload),
                    )
                    .await?;
                Ok(res.1)
```

sdk/rust/src/dstack_client.rs
```diff
@@ -0,0 +1,272 @@
...
+    }
+
+    pub async fn info(&self) -> Result<InfoResponse> {
+        let response = self.send_rpc_request("/Info", &json!({})).await?;
+        Ok(InfoResponse::validated_from_value(response)?)
+    }
+
+    pub async fn emit_event(&self, event: String, payload: Vec<u8>) -> Result<()> {
+        if event.is_empty() {
+            return Err(anyhow!("Event name cannot be empty"));
```
kvinwang:
```suggestion
            anyhow::bail!("Event name cannot be empty");
```

sdk/rust/src/dstack_client.rs
tuddman: This reads like ai wrote it.  
the struct `TlsKeyConfig` doesn't exist!
What fields would you intend to add?  this is for retrieving a TLS Key, which from the rpc request expects a regular json payload, with the stable fields, that is then deserialized as a `GetTlsKeyResponse` anyway.

kvinwang: ## Background
Currently, dstack only supports a few options for TLS certificates in this API. It is very likely that more options will be added in future versions.

## The problem
If we keep the current Rust `get_tls_key` API style, each time we add another parameter to this API in future versions, all users of this function will encounter compilation errors. This is not desirable.

For example, users call this function with the current version of the SDK:
```rust
let _k = client.get_tls_key(None, None, true, true, true).await?;
```
Then, if we add one more parameter like `enable_spiffe` to the function in a future version, SDK users would have to update their code to:
```rust
let _k = client.get_tls_key(None, None, true, true, true, false).await?;
```
This means each time we introduce a new parameter, it would create a breaking change. Additionally, the code at the call site becomes less readable.

## Solution
We can define the arguments as a struct with builder pattern enabled, as shown below:
```rust
#[bon::Builder]
struct TlsKeyConfig {
        #[builder(into)]
        subject: Option<String>,
        alt_names: Option<Vec<String>>,
        #[builder(default = false)]
        usage_ra_tls: bool,
        #[builder(default = true)]
        usage_server_auth: bool,
        #[builder(default = false)]
        usage_client_auth: bool,
}
pub async fn get_tls_key(&self, config: TlsKeyConfig)
```

The caller code would become:
```rust
let _k = client.get_tls_key(
        TlsKeyConfig::builder()
             .subject("MyServer")
             .alt_names(vec![])
             .usage_ra_tls(true)
             .usage_server_auth(false)
             .usage_client_auth(true)
             .build()
).await?;
```

It can also omit some fields and use the default values:
```rust
let _k = client.get_tls_key(
        TlsKeyConfig::builder()
             .subject("MyServer")
             .usage_ra_tls(true)
             .build()
).await?;
```
You can see this is more readable and less error-prone.

When we add more fields in a future version:
```rust
#[bon::Builder]
struct TlsKeyConfig {
        #[builder(into)]
        subject: Option<String>,
        alt_names: Option<Vec<String>>,
        #[builder(default = false)]
        usage_ra_tls: bool,
        #[builder(default = true)]
        usage_server_auth: bool,
        #[builder(default = false)]
        usage_client_auth: bool,
        #[builder(default = false)]
        enable_spiffe: bool,
}
pub async fn get_tls_key(&self, config: TlsKeyConfig)
```
The previous calling statements would still be valid, so SDK users wouldn't need to update their code.

tuddman: added in [31a8db1](https://github.com/Dstack-TEE/dstack/pull/161/commits/31a8db16bf546836ce7fd82612f3359515b103a1)

kvinwang:
Thank you so much for your valuable contribution!

PR #175, opened 2025-05-20
kms: Implements os_image_hash verification

kvinwang:
This PR implements mr_image verification in KMS. So that we can register a simple mr_image on-chain to validate the CVM image.
Note the GPU CVM is not supported yet, as it's MRs would be a bit complicated.


Leechael:
lgtm

PR #182, opened 2025-05-24
KMS: add factory deployment for AppAuth contracts with single-transaction optimization

Leechael:


kms/auth-eth/contracts/KmsAuth.sol
```diff
@@ -120,6 +126,51 @@ contract KmsAuth is
...
+        // Deploy proxy contract
+        proxyAddress = address(new ERC1967Proxy(appAuthImplementation, initData));
+        
+        // Register to KMS
+        require(!apps[appId].isRegistered, "App already registered");
+        apps[appId].isRegistered = true;
+        apps[appId].controller = proxyAddress;
+        nextAppSequence[msg.sender]++;
+        
+        emit AppRegistered(appId);
```
kvinwang: It looks like this part is a duplicate of the registerApp logic. Consider extracting the shared functionality into _registerApp.

kms/auth-eth/contracts/KmsAuth.sol
```diff
@@ -120,6 +126,51 @@ contract KmsAuth is
...
+            initialOwner,
+            appId,
+            disableUpgrades,
+            allowAnyDevice,
+            initialDeviceId,
+            initialComposeHash
+        );
+        
+        // Deploy proxy contract
+        proxyAddress = address(new ERC1967Proxy(appAuthImplementation, initData));
```
h4x3rotab: I need to double check if this is the preferred way to deploy a UUPS upgradable contract.

kms/auth-eth/contracts/KmsAuth.sol
h4x3rotab: Agree, better to move the shared code to a private function

kms/auth-eth/contracts/KmsAuth.sol
```diff
@@ -48,6 +49,9 @@ contract KmsAuth is
     // Sequence number for app IDs - per user
     mapping(address => uint256) public nextAppSequence;
 
+    // AppAuth implementation contract address for factory deployment
+    address public appAuthImplementation;
```
h4x3rotab: I suggest to make it initialized during construction function. In case we need to change it we can run an upgrade later. But it's ok to consider it in future version.

kms/auth-eth/contracts/AppAuth.sol
```diff
@@ -47,11 +47,63 @@ contract AppAuth is
...
+            _appId,
+            _disableUpgrades,
+            _allowAnyDevice,
+            bytes32(0),  // no initial device
+            bytes32(0)   // no initial hash
+        );
+    }
+
+    // Initialize the contract with initial device and compose hash
+    function initializeWithData(
```
h4x3rotab: I suggest to merge it to the initialize() function. We can address that by setting the default initial device and initial hash to zero.

kvinwang (approved):
LGTM

PR #207, opened 2025-06-11
gw: Seperate proxy runtime from Rocket

kvinwang:


Cargo.toml
```diff
@@ -87,6 +87,7 @@ tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
 safe-write = "0.1.2"
 nix = "0.29.0"
 sd-notify = "0.4.5"
+jemallocator = "0.5.4"
```
Leechael, line 90: [jemalloc](https://github.com/jemalloc/jemalloc) has archived. Should we consider another option?

kvinwang: I know it's archived. But it's not necessary to be scared. It is mature, excellent, and is widely used in world production softwares. If one day it encounters a critical CVE which affects our product, and no one fixes it, then we can easily remove it by removing one line of code. It's even easier than upgrading a dependency.

kvinwang: Seems it is actually [maintained by facebook](https://github.com/facebook/jemalloc/discussions/7) for years😄

PR #223, opened 2025-06-23
feat(sdk/js): backward compatibility for smooth migration

Leechael:
This PR proposes releasing a beta version of the JS SDK that maintains backward compatibility. Users still need to bind `/var/run/dstack.sock` during deployment, but it aims to minimize code changes in users' existing code.

To learn more about renaming, check https://github.com/Dstack-TEE/dstack/pull/142

### DIFF
--- a/sdk/js/package.json
+++ b/sdk/js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@phala/dstack-sdk",
-  "version": "0.5.0-beta.2",
+  "version": "0.5.0",
   "description": "Dstack SDK",
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/sdk/js/src/send-rpc-request.ts
+++ b/sdk/js/src/send-rpc-request.ts
@@ -2,7 +2,7 @@ import http from 'http'
 import https from 'https'
 import net from 'net'
 
-export const __version__ = "0.5.0-beta.1"
+export const __version__ = "0.5.0"
 
 
 export function send_rpc_request<T = any>(endpoint: string, path: string, payload: string, timeoutMs?: number): Promise<T> {


sdk/js/src/get-compose-hash.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+function dumpAppCompose(dic: AppCompose): string {
+  const ordered = sortObject(dic);
+  let json = JSON.stringify(ordered, null, 4);
+  json = json.replace(/": /g, '":');
+  return json;
+}
+
+export function getComposeHash(app_compose: AppCompose): string {
```
h4x3rotab: It's user's responsibility to normalize app-compose file, not ours. So here we cannot force reprocess the app compose file before returning the hash. Instead, we should only give a function "normalizeAppCompose" and let the user to choose if they apply it or not

sdk/js/src/get-compose-hash.ts
```diff
@@ -0,0 +1,58 @@
...
+  if ("pre_launch_script" in obj && !obj.pre_launch_script) {
+    delete obj.pre_launch_script;
+  }
+  return obj;
+}
+
+function dumpAppCompose(dic: AppCompose): string {
+  const ordered = sortObject(dic);
+  let json = JSON.stringify(ordered, null, 4);
+  json = json.replace(/": /g, '":');
```
h4x3rotab: Dangerous to do this. Imagine you have `": ` in the prelaunch script, it will corrupt the file

sdk/js/src/get-compose-hash.ts
Leechael: The `preprocessAppCompose` is now configurable, but `toDeterministicJson` is still kept for consistent results across different programming languages.

docs/determinstic-compose-hash.md
```diff
@@ -0,0 +1,246 @@
+# Deterministic Compose Hash
```
h4x3rotab: I suggest to call it "Normalized App Compose"

PR #228, opened 2025-07-02
ACPI Tables Dynamic Generation

kvinwang:
This PR reimplements the `dstack-mr` in Rust and generates the ACPI tables dynamically according to the HW config rather than embedding a bunch of templates in the executable binary. This enables the KMS verification for more flexible HW variants such as GPU enabled CVMs.
The corresponding command `dstack-acpi-tables` is a modified version of QEMU and can be built from:
https://github.com/kvinwang/qemu-tdx/commit/d98440811192c08eafc07c7af110593c6b3758ff


jkr0103:
@kvinwang I was trying to use the qemu patch for ACPI dump

https://github.com/kvinwang/qemu-tdx/commit/d98440811192c08eafc07c7af110593c6b3758ff

I see ACPI tables dumped just after construction differ with what we see inside a TD, changes to the above patch seems to remove that difference during table construction itself.

Could you please help me to understand the difference?

h4x3rotab:
@jkr0103 Hey, I've sent you an email (the intel address). Happy to follow up there.

PR #230, opened 2025-07-02
gateway: Add api for evidences

kvinwang:
This PR adds two APIs to the gateway that exposes necessary information to verify its dstack TEE authenticity.
The endpoints are:
- `https://gateway.<zt-https-base-domain>/app-info` returns the app info got from the dstack-guest-agent
- `https://gateway.<zt-https-base-domain>/acme-info` returns the zt-https certs and it's quote+eventlog

### DIFF
--- a/gateway/src/proxy/tls_terminate.rs
+++ b/gateway/src/proxy/tls_terminate.rs
@@ -13,6 +13,7 @@ use hyper_util::rt::tokio::TokioIo;
 use rustls::pki_types::pem::PemObject;
 use rustls::pki_types::{CertificateDer, PrivateKeyDer};
 use rustls::version::{TLS12, TLS13};
+use serde::Serialize;
 use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
 use tokio::net::TcpStream;
 use tokio::time::timeout;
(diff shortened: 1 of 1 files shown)


Leechael:
1. Suggesting to add `.well-known` prefix and `dstack` prefix, for example:
- `https://gateway.<zt-https-base-domain>/.well-known/dstack-app-info`
- `https://gateway.<zt-https-base-domain>/.well-known/dstack-acme-info`

`.well-known` is a path prefix defined in [rfc8615](https://datatracker.ietf.org/doc/html/rfc8615) and [wikipedia](https://en.wikipedia.org/wiki/Well-known_URI)

2. Also suggesting support via instance ID:
- `https://<instance_id>.<zt-https-base-domain>/.well-known/dstack-{app-info,acme-info}`, which would avoid exposing the built-in guest-api via port 8090 that contains sensitive information

kvinwang:
> 1. Suggesting to add `.well-known` prefix and `dstack` prefix, for example:
> 
> * `https://gateway.<zt-https-base-domain>/.well-known/dstack-app-info`
> * `https://gateway.<zt-https-base-domain>/.well-known/dstack-acme-info`
> 
> `.well-known` is a path prefix defined in [rfc8615](https://datatracker.ietf.org/doc/html/rfc8615) and [wikipedia](https://en.wikipedia.org/wiki/Well-known_URI)

I've created aliases for the entries under `/.dstack` as a potential convention for future use.

PR #231, opened 2025-07-02
KmsAuth: Use the contract address as app-id

kvinwang:
This PR uses the app contract address as the app-id. This simplifies the code logic and reduces one id, making it easier to understand.


kms/auth-eth/contracts/AppAuth.sol
```diff
@@ -131,11 +125,6 @@ contract AppAuth is
     function isAppAllowed(
         IAppAuth.AppBootInfo calldata bootInfo
     ) external view override returns (bool isAllowed, string memory reason) {
-        // Check if this controller is responsible for the app
-        if (bootInfo.appId != appId) {
-            return (false, "Wrong app controller");
-        }
-
```
Leechael, line 138: We also need to remove `appId` from `AppBootInfo`?

https://github.com/Dstack-TEE/dstack/blob/e26d73d9bf7c3ef8010395cc7220187a2b3fe748/kms/auth-eth/contracts/IAppAuth.sol#L32

kms/auth-eth/contracts/KmsAuth.sol
```diff
@@ -30,15 +30,6 @@ contract KmsAuth is
...
     string public gatewayAppId;
 
-    // Struct to store App configuration
-    struct AppConfig {
-        bool isRegistered;
-        address controller;
-    }
-
-    // Mapping of registered apps
-    mapping(address => AppConfig) public apps;
```
h4x3rotab, line 40: I think it's still necessary to have a map to save which app contract is connected to this KMS. We can consider to make it a `mapping(address => bool)`

kms/auth-eth/contracts/AppAuth.sol
kvinwang: It's used in KmsAuth.sol.

kms/auth-eth/contracts/KmsAuth.sol
kvinwang: OK

Leechael (approved):
lgtm

Issue #214, opened 2025-06-17
Naming and concepts around onchain KMS


h4x3rotab:
While onchain KMS is already widely used in our daily discussion, it's more an implementation detail instead of a good user facing term. Onchain KMS refers to a decentralized TEE app governance architecture, where the KMS is aware of the smart contract (`AppAuth`) deployed to control how the TEE app can be deployed and upgraded. On the other hand, the KMS itself is also a special contract managing the KMS app (also an app running on dstack) itself.

To add some background, here's a tl;dr of the onchain KMS architecture:

- From the latest version, dstack has the onchain KMS capability. It can be turn off with "dev mode".
- A full stack dstack includes `dstack-vmm`, `dstack-kms`, and `dstack-gateway`
   - `dstack-kms` and `dstack-gateway` must be paired
   - A `dstack-kms` is controlled by a `KmsAuth` smart contract (on a specific chain)
   - Both `dstack-kms` and `dstack-gateway` can have multiple instance for redundancy
- All the CVMs deployed on this node:
   - Should use the corresponding `dstack-kms` and `dstack-gateway`
   - Should specify the address of the corresponding `AppAuth` contract (living on the same chain as `KmsAuth` contract)
   - Should have its current whitelisted versions set in its `AppAuth` contract

Below are a list of the related names and concepts. I will describe what it means, how it's used, and the new candidates. In this issue, we request comments and ideas.

## Node type - Onchain KMS

It refers to a dstack deployment (dstack node) with onchain KMS enabled, to distinguish it from non onchain kms enabled dstack node. This name can be used when showing the node type from the CVM get-info API or when choosing the node on the Phala Cloud.

| Node type on CVM card  | Node type in the node choice box  |
|---|---|
| <img width="300" alt="Image" src="https://github.com/user-attachments/assets/df8e9d85-5226-47a7-ad5d-e5e5900e0771" /> | <img width="240" alt="Image" src="https://github.com/user-attachments/assets/d6423def-fe8b-4c68-b388-6cc26de9d220" /> |

Ideally we should focus on what features or benefits it brings to the user, instead of the technical details of the node.

Candidates:

- Onchain Gov: Address the app upgrade is managed by onchain smart contracts
   - Chain Gov
   - Chain Managed
- Contract Owned: Address the app is not controlled by an entity, but a smart contract
   - Contract Gov
- DAO Owned: Address the app can be owned by a DAO (but actually not necessarily true as it can be an EOA)
   - DAO Gov
[...]

PR #240, opened 2025-07-14
kms: Rename contracts

kvinwang:
This PR rename the solidity contracts according to https://github.com/Dstack-TEE/dstack/issues/214#issuecomment-3067852102


kms/auth-eth/typechain-types/contracts/index.ts
```diff
@@ -1,7 +1,9 @@
 /* Autogenerated file. Do not edit manually. */
 /* tslint:disable */
 /* eslint-disable */
+export type { AppAuth } from "./AppAuth";
```
kvinwang: @Leechael Why do you add the old name back?

PR #241, opened 2025-07-15
Implements CVM disk backup

kvinwang:
This PR implements CVM disk backup in both built-in vmm and external script `dstack-backup.py`.
As the backup usually takes a very long time, **it is recommended to run with the external script `dstack-backup.py`** rather than vmm RPC.
How to use:
1. Install dependency [qmpbackup](https://github.com/kvinwang/qmpbackup)
2. Enable qmp socket in vmm.toml
3. Add `dstack-backup.py` to crontab or similar task scheduler with a daily trigger. For example:
   ```
   30 2 * * * /path/to/dstack-backup.sh
   ```
    where dstack-backup.sh can be written as:
    ```bash
    #!/bin/bash
    cd /opt/meta-dstack/build
    source .venv/bin/activate
    # use flock to prevent reentrant
    flock -n ./backup.lck ./dstack-backup.py --vmm-work-dir . --max-backups 3 --full-interval 7d --inc-interval 1d
    ```
It will backup the disks of running VMs to a configured directory.
Run `dstack-backup.py --help` to show the arguments.
There are two kinds of backups, `full` and `incremental`. By default, it performs full backup weekly, and incremental backup daily.


guest-agent/src/guest_api_service.rs
```diff
@@ -112,6 +115,53 @@ impl GuestApiRpc for GuestApiHandler {
...
     }
+
+    async fn pre_backup(self) -> Result<()> {
+        fs::OpenOptions::new()
+            .create_new(true)
+            .write(true)
+            .open(BACKUP_LOCK_FILE)
+            .context("Failed to create backup lock file, there is another backup in progress")?;
+        // Run /dstack/hooks/pre-backup if it exists
+        let pre_backup_hook = "/dstack/hooks/pre-backup";
```
Leechael, line 126: How can we set up hooks for pre-backup and post-backup?

vmm/rpc/proto/vmm_rpc.proto
```diff
@@ -224,6 +224,41 @@ message GpuInfo {
   bool is_free = 4;
 }
 
+message BackupDiskRequest {
+  // vm id
+  string vm_id = 1;
+  // full or incremental
+  string level = 2;
```
Leechael, line 232: Is it better to use Enum for readability?

vmm/src/app.rs
```diff
@@ -653,6 +668,232 @@ impl App {
...
+                let timestamp = chrono::Utc::now().format("%Y%m%dZ%H%M%S").to_string();
+                let new_dir = backup_dir.join(&timestamp);
+                fs::create_dir_all(&new_dir).context("Failed to create backup directory")?;
+                if fs::symlink_metadata(&latest_dir).is_ok() {
+                    fs::remove_file(&latest_dir)
+                        .context("Failed to remove latest directory link")?;
+                }
+                fs::os::unix::fs::symlink(&timestamp, &latest_dir)
+                    .context("Failed to create latest directory link")?;
+            }
```
Leechael, line 710: I'm suggesting to add a checksum for each backup.

vmm/src/main_service.rs
```diff
@@ -461,6 +492,34 @@ impl VmmRpc for RpcHandler {
         let hash = hex_sha256(&request.compose_file);
         Ok(RpcComposeHash { hash })
     }
+
+    async fn backup_disk(self, request: BackupDiskRequest) -> Result<()> {
+        check_path_traversal(&request.vm_id)?;
+        self.app.backup_disk(&request.vm_id, &request.level).await
```
Leechael, line 528: Can we show the backup progress in `VmInfo` too?

guest-agent/src/guest_api_service.rs
kvinwang: Just write some shell commands into the files `pre-backup/post-backup`. It usaully need nothing todo. Some apps may want to flush their app data to disk before backup.

vmm/rpc/proto/vmm_rpc.proto
kvinwang: It's lack of syntax in the proto to express `#[serde(rename = "snake_case")]` when we convert it to json API.
So, I usually avoid to use enum.

vmm/src/app.rs
kvinwang: Good point. The feature would be better to add to the qmpbackup command, since it controls the filename generation.

vmm/src/main_service.rs
kvinwang: Adding pregress value might not trival. May add a boolean flag indicating whether backup in progress.

guest-agent/src/guest_api_service.rs
Leechael: Are they set up only by node operators? I'm wondering who can actually create these hook scripts.

vmm/src/app.rs
Leechael: Even better: can we verify the checksum before restoring?

kvinwang: > Even better: can we verify the checksum before restoring?

Surely should add verification in `qmprestore`, if checksum is added to `qmpbackup`.

guest-agent/src/guest_api_service.rs
kvinwang: > Are they set up only by node operators? I'm wondering who can actually create these hook scripts.

It's up to app defines the logic in the hooks. For example, flush their mysql database in the CVM. The hooks is put there for future use. I think we don't need to care about it at this time, until their is a use case appears.

vmm/src/main_service.rs
kvinwang: Added in https://github.com/Dstack-TEE/dstack/pull/241/commits/30bbfa082aacdff030f2d0dce579e167b72f381a

PR #259, opened 2025-07-28
Make KMS configurable

ImTei:
Hi! I'm Tei from Sunnyside Labs.

This PR adds configuration options for running KMS using docker compose and run script.
- `VERIFY_IMAGE`: Allow disabling image hash verification feature to use custom dstack OS image.
- `GIT_REPOSITORY`: Allow to compile and run KMS from forked repository.

I had to disable `verify` option to run my custom dstack OS image. And to apply & test this change, I needed to specify my forked repository in the docker compose file.
This change is backward compatible and does not affect existing behavior unless configured explicitly.

### DIFF
--- a/kms/dstack-app/deploy-to-vmm.sh
+++ b/kms/dstack-app/deploy-to-vmm.sh
@@ -32,8 +32,14 @@ else
 # The URL of the dstack app image download URL
 # IMAGE_DOWNLOAD_URL=https://files.kvin.wang/images/mr_{OS_IMAGE_HASH}.tar.gz
 
+# Image hash verification feature flag
+# VERIFY_IMAGE=true
+
 # The URL of the Ethereum RPC service
 ETH_RPC_URL=https://rpc.phala.network
+
+# The Git repository to deploy
+# GIT_REPOSITORY=https://github.com/Dstack-TEE/dstack.git
 
 # The Git revision to deploy
 GIT_REV=HEAD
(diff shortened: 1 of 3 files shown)


kms/dstack-app/deploy-to-vmm.sh
```diff
@@ -32,9 +32,15 @@ else
 # The URL of the dstack app image download URL
 # IMAGE_DOWNLOAD_URL=https://files.kvin.wang/images/mr_{OS_IMAGE_HASH}.tar.gz
 
+# Image hash verification feature flag
+# VERIFY_IMAGE=true
```
kvinwang: Let's uncomment them for backward compatibility.
```suggestion
VERIFY_IMAGE=true
```

kms/dstack-app/deploy-to-vmm.sh
```diff
@@ -32,9 +32,15 @@ else
...
+
 # The URL of the Ethereum RPC service
 ETH_RPC_URL=https://rpc.phala.network
 
+# The Git repository to deploy
+# GIT_REPOSITORY=https://github.com/Dstack-TEE/dstack.git
```
kvinwang:
```suggestion
GIT_REPOSITORY=https://github.com/Dstack-TEE/dstack.git
```

kvinwang (commented):
Thanks!

kvinwang:
> But it seems this compose file does not use inline dockerfile anymore. so we can't use `GIT_REPOSITORY`.
> Since the purpose of `GIT_REPOSITORY` is for applying `VERIFY_IMAGE`, I think we can just remove `GIT_REPOSITORY` and merge only `VERIFY_IMAGE`. wdyt?

Sounds good to me. The `VERIFY_IMAGE` configuration would be valuable for most developers. For additional customization, it would be preferable to create their own docker compose file.

ImTei:
oh just noticed that auth-api container is still using an inline dockerfile 😅
left `GIT_REPOSITORY` just in case. and fixed merge conflict @kvinwang

kvinwang (approved):

PR #260, opened 2025-07-28
Make certbot cli support multi domains

ImTei:
Hi! I'm Tei from Sunnyside Labs.

This PR enables certbot CLI to manage TLS cert of multi domain names.
This change is a breaking change, and will break the existing cerbot configs.


kvinwang (commented):
Thanks, LGTM.
This is a breaking change to the config of standalone `certbot`. @Leechael , do you still use it?

PR #272, opened 2025-08-04
refactor: break up `dstack-sdk` into two crates, client and types

### LINKED ISSUE #271: Make Dstack Rust SDK `no_std` and I/O-free for smart contract compatibility
### TL;DR

I think it would it make sense to make the [Dstack Rust SDK](https://github.com/Dstack-TEE/dstack/tree/master/sdk/rust) crate `no_std` and avoid any I/O dependencies, so we can use it in a smart contract context.

### Context

I ran into an issue where I added a `near-sdk` dependency to a crate that's part of a NEAR smart contract, and it failed to compile because `near-sdk` pulls in `tokio`, which in turn [pulls in](https://github.com/near/mpc/actions/runs/16677660995/job/47208197879?pr=758) `mio`. As a workaround, I just [copy-pasted](https://github.com/near/mpc/blob/9c70939f52bd86f6232842539193b5ad4d215400/attestation/src/attestation.rs#L16-L35) [the parts](https://github.com/near/mpc/blob/9c70939f52bd86f6232842539193b5ad4d215400/attestation/src/tcbinfo.rs#L7-L37) I needed from the `near-sdk` repo, but I'd much prefer to import `near-sdk` directly — it’s cleaner and more maintainable that way.

DSharifi:
closes #271 

No new functionality is added to the code in this PR. It is a refactor that breaks out the types in `dstack-sdk` into a separate crate such that:
1. `dstack-sdk`. This is the client crate and contains the logic of communicating with the `dstack` socket. 
2. `dstack-sdk-types` Contains all the public types used for communicating with the dstack process, such as `EventLog`, and types that external users of the SDK depend on verify remote attestations. This crate is `I/O` free and also has `no_std` enabled to make it compatible for smart contracts.

To enforce that `dstack-sdk-types` maintains WebAssembly compatibility, a step has been added to CI.

The changes are fully backwards compatible, as all the types that were previously in `dstack-sdk` and now in `dstack-sdk-types` are publicly re-exported.


sdk/rust/clients/Cargo.toml
```diff
@@ -1,5 +1,5 @@
 [package]
-name = "dstack-sdk"
+name = "dstack-sdk-clients"
```
kvinwang: Thank you for the contribution. I'd like to keep the crate name `dstack-sdk-clients` as `dstack-sdk`, for compatibility with other users and consistency with other language implementations.

PR #285, opened 2025-08-11
feat: implement `Debug` for all public rust SDK types

DSharifi:
`Debug` shoul as a best practice be implemented by all public API types for a library.


netrome:
As an extension on this, please consider eagerly implementing all common traits as per https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits, so we don't have to follow-up with dedicated PRs as soon as we'd need `PartialEq`, `Clone`, `Hash` or any other common trait on these types.

PR #301, opened 2025-08-18
update SPDX annotations

DSharifi:


sdk/rust/types/Cargo.toml
```diff
@@ -28,6 +28,7 @@ tokio = { workspace = true, features = ["full"] }
 [features]
 default = ["std"]
 borsh = ["dep:borsh"]
+borsh_schema = ["borsh", "borsh/unstable__schema"]
```
kvinwang: It shouldn't be in this PR.

sdk/rust/types/Cargo.toml
```diff
@@ -36,5 +37,5 @@ std = [
     "serde/std",
     "serde_json/std",
     "sha2/std",
-    "borsh?/std"
+    "borsh?/std",
```
kvinwang: ditto

sdk/rust/types/Cargo.toml
DSharifi: Good catch, seems like I forgot to branch from main. I'll update.

PR #303, opened 2025-08-19
certbot: Add detailed acme error info

kvinwang:


certbot/src/dns01_client/cloudflare.rs
```diff
@@ -83,7 +83,6 @@ impl Dns01Api for CloudflareClient {
                 "type": "TXT",
                 "name": domain,
                 "content": content,
-                "ttl": 120
```
Leechael, line 86: What's the default value if we remove `ttl`?

kvinwang: Cloudflare says it is default is 300; however, my test shows it is faster than 120, appearing to be around 60.

PR #306, opened 2025-08-19
rdme: Update access for non-kms apps

kvinwang:
Updates README documentation with details on non-kms application connectivity to agent dashboards and services, featuring gRPC access protocols via the zt-https proxy gateway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,20 +183,12 @@ Open the dstack-vmm webpage [http://localhost:9080](http://localhost:9080) (chan
 
 After the container is deployed, it should take some time to start the CVM and the containers. Time would be vary depending on your workload.
 
-- **[Logs]**: Click the button to see the logs of the CVM, you can see if the container is finished starting there
-- **[Dashboard]**: Once the container is running, you can click the button to see some information about the container. The logs of the containers can be seen in the Dashboard page
+- **[Logs]**: Click this button to view the CVM logs and monitor container startup progress
+- **[Dashboard]**: Once the container is running, click this button to view container information and logs. (Note: This button is only visible when dstack-gateway is enabled. If disabled, you'll need to add a port mapping to port 8090 to access the CVM dashboard)
 
 <div align="center">
 
 <img src="./docs/assets/guest-agent.png" alt="Guest Agent Dashboard" height="300">
-
-</div>
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -214,19 +206,24 @@ The environment variables will be encrypted in the client-side and decrypted in
...
+**Examples**:
 
-- `<id>[s].<base_domain>` maps to port `80` or `443` if with `s` in the CVM
-- `<id>-<port>[s].<base_domain>` maps to port `<port>` in the CVM
+- `3327603e03f5bd1f830812ca4a789277fc31f577-8080.app.kvin.wang` → port `8080` (HTTP/1.1)
+- `3327603e03f5bd1f830812ca4a789277fc31f577-8080g.app.kvin.wang` → port `8080` (HTTPS/2 with TLS termination)
+- `3327603e03f5bd1f830812ca4a789277fc31f577-8080s.app.kvin.wang` → port `8080` (TLS passthrough)
 
-**Example**: `3327603e03f5bd1f830812ca4a789277fc31f577-8080.app.kvin.wang` maps to port `8080` in the CVM.
+The `<id>` can be either the app ID or instance ID. When using the app ID, the load balancer will select one of the available instances. Adding an `s` suffix enables TLS passthrough to the app instead of terminating at dstack-gateway. Adding a `g` suffix enables HTTPS/2 with TLS termination for gRPC applications.
```
h4x3rotab, line 219: Two questions:

- Is it only for gRPC, or also used for other http/2 applications?
- Is it necessary to have the `g` suffix?

kvinwang: - It's for all HTTP/2.
- Without `g`, the gateway doesn't know HTTP/2 backend support. Falsely advertising HTTP/2 compatibility would result in the client sending incompatible packets to an HTTP/1.1 server.

h4x3rotab: Do you think it's better to use `-h2` instead of `-g`? And for custom domain, does it detect the suffix in the TXT record?

kvinwang: `*-8080-h2` would be a breaking change introducing another compatibility issue, `*-8080h2` looks a bit weird to me.

> And for custom domain, does it detect the suffix in the TXT record?

No, custom domain knows itself support h2 or not, it can only advertising h2 when the backend support it.

PR #311, opened 2025-08-27
fix: Dockerfile for gateway dstack app build

Leechael:


gateway/dstack-app/Dockerfile
```diff
@@ -27,7 +27,7 @@ WORKDIR /src/dstack
 RUN cargo build --release -p dstack-gateway
 
 # Runtime stage
-FROM debian:bookworm@sha256:ced9eb5eca0a3ba2e29d0045513863b3baaee71cd8c2eed403c9f7d3eaccfd2b
+FROM debian:bookworm@sha256:731dd1380d6a8d170a695dbeb17fe0eade0e1c29f654cf0a3a07f372191c3f4b
```
kvinwang, line 30: This Dockerfile should be removed entirely since we've added the reproducible builder. Use the docker image on the hub instead, in the docker-compose.yaml. **Sorry for the incomplete work I did**.

PR #348, opened 2025-09-26
cvm: Support for ext4

kvinwang:
This PR allows the app to select using zfs or ext4 as the storage filesystem.

### Caution
**ZFS:** Ensures filesystem integrity with built-in data protection features.
**ext4:** Provides better performance for database applications with lower overhead and faster I/O operations, but no strong integrity protection.


kvinwang:
> We should update the guest agent to expose the current storage filesystem.

It's already in the app-compose.json. Also put in the RTMR3 event log.

PR #356, opened 2025-09-28
feat(gateway): added proxy protocol support.

Leechael:


Leechael:
Demo implementation for discussion, closed because we can out another implementation plan.

PR #358, opened 2025-09-28
cvm: Add built-in swap config

kvinwang:


Leechael:
> > We should support the ability to update swap size before shipping it, because only allowing creation without the option to adjust swap size will seem strange to users.
> 
> They can update the compose file and reboot the CVM. It is just not adjustable after booting at runtime.

Does this mean the swap will always be recreated when the CVM reboots, and the mount configuration won't persist?

kvinwang:
> Does this mean the swap will always be recreated when the CVM reboots, and the mount configuration won't persist?

The swap data is on the persistent disk, but it was always removed and create a new one on each boot.
https://github.com/Dstack-TEE/dstack/pull/358/files#diff-d6725e0bb1459ef41e5975c6f4cd01db7c81d5b76297f4632b934adbbabe4729R699

Leechael (approved):

PR #376, opened 2025-10-21
fix(kms): package name

chuwt:
package name in commit [78057c975fe4b9e21f557fb888d72eeecfb21178](https://github.com/Dstack-TEE/dstack/blob/78057c975fe4b9e21f557fb888d72eeecfb21178/kms/Cargo.toml) is `kms`.

### DIFF
--- a/kms/dstack-app/docker-compose.yaml
+++ b/kms/dstack-app/docker-compose.yaml
@@ -73,8 +73,8 @@ services:
             cd dstack && \
             git checkout 78057c975fe4b9e21f557fb888d72eeecfb21178
         WORKDIR /app/dstack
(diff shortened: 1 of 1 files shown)


kvinwang:
It was renamed to `dstack-kms` months ago.

chuwt:
Yeah, but the current deployment is still using the previous codebase:
https://github.com/Dstack-TEE/dstack/blob/4631d945998047cd920b20342aaefcdff67f5c95/kms/dstack-app/docker-compose.yaml#L74-L77

kvinwang:
> Yeah, but the current deployment is still using the previous codebase:

Thanks for catching that. I think the git commit should be updated.

PR #387, opened 2025-11-05
vmm: Cache guest events in a buffer of size 20

kvinwang:
For a better status observation ability.


vmm/src/app.rs
```diff
@@ -411,6 +411,13 @@ impl App {
         let Some(vm) = state.vms.values_mut().find(|vm| vm.config.cid == cid) else {
             bail!("VM not found");
         };
+        vm.state.events.push_back(pb::GuestEvent {
+            event: event.into(),
+            body: body.clone(),
+        });
```
Leechael, line 422: Can we add a timestamp too?

kvinwang: > Can we add a timestamp too?

Sure, added.

Leechael (approved):

PR #436, opened 2026-01-14
Make DNS TXT record TTL configurable.

Evrard-Nil:
Default setting by cloudflare is 300s which can be slow.


kvinwang:
Thanks! Let's add a configuration item here as well: https://github.com/Dstack-TEE/dstack/blob/b9142141964e14fffc2e2dbcd7207f51526419ab/gateway/dstack-app/deploy-to-vmm.sh#L84-L85

PR #506, opened 2026-02-13
docs: add encrypted env vars technical spec

Leechael:
## Summary
- Adds `docs/encrypted-env-spec.md` documenting the X25519 + AES-256-GCM scheme used to encrypt/decrypt application environment variables
- Covers encryption public key source (KMS derivation chain, RPC interface, signature verification)
- Documents the `.appkeys.json` file format and hex-encoded field conventions
- Provides cross-language implementation guides for Rust, Go, Python, and TypeScript (both encrypt and decrypt)
- Includes end-to-end flow diagram from deploy-time encryption through TEE-side decryption

## Updates after review
- Added **Security Considerations** section to explicitly state that encrypted env provides confidentiality but **not origin authentication**
- Added developer guidance for application-layer authenticity checks:
  1. `APP_LAUNCH_TOKEN` pattern
  2. custom signature verification
  3. embedded shared secret verification
- Added links to:
  - `docs/security/security-best-practices.md#authenticated-envs-and-user_config`
  - `docs/security/security-model.md#environment-variables-need-application-layer-authentication`
- Added `user_config` parallel caveat (same authenticity/integrity risk class)
- Clarified that `app_id` must be computed from the normalized `app-compose.json` bytes used by dstack compose hashing
- Added **Runtime File/Path Contract (dstack)** section to avoid treating runtime file names/locations as arbitrary:
  - `/dstack/.host-shared/app-compose.json`
  - `/dstack/.host-shared/.encrypted-env`
  - `/dstack/.host-shared/.appkeys.json`
  - `/dstack/.host-shared/.decrypted-env`
  - `/dstack/.host-shared/.decrypted-env.json`

## Commits in this update
- `9e886cbc` docs: add security considerations for encrypted env authenticity
- `93728db1` docs: clarify normalized app_id input and runtime file path contract

## Test plan
- [x] Review doc accuracy against source code (`dstack-util/src/crypto.rs`, `kms/src/main_service.rs`, SDK implementations)
- [x] Verify consistency with existing security docs (`security-model.md`, `security-best-practices.md`)
- [x] Confirm review concern is explicitly addressed in the spec text

### DIFF
--- a/docs/encrypted-env-spec.md
+++ b/docs/encrypted-env-spec.md
@@ -1,0 +1,444 @@
+# Encrypted Environment Variables
+
+dstack uses an ECIES variant (X25519 + AES-256-GCM) to protect application environment variables. The client encrypts env vars with an X25519 public key at deploy time. At boot, the CVM obtains the corresponding private key from KMS via TDX remote attestation and decrypts inside the TEE.
+
+## Encryption Public Key Source
+
+### Key Derivation Chain
+
+The KMS deterministically derives a per-application key pair from its root CA key:
+
+```
+KMS root CA key (P-256 KeyPair)
+  │
+  └─ derive_dh_secret(context = [app_id, "env-encrypt-key"])
+       → SHA256(derived_P256_key_DER) → 32 bytes
+       → X25519 StaticSecret (private key = env_crypt_key, delivered to TEE)
+       → X25519 PublicKey (public key, exposed to client for encryption)
+```
+
+The same `app_id` always derives the same key pair.
+
+### Computing `app_id`
+
+```
+app_id = SHA256(app-compose.json)[0..20]    // first 20 bytes, 40 hex characters
+```
+
+Example:
+
+```javascript
+const composeHash = sha256(composeJsonString);   // 32 bytes hex
+const appId = composeHash.slice(0, 40);          // first 20 bytes = 40 hex chars
+```
+
+### RPC Interface
+
+The public key is exposed through a two-level RPC chain:
+
+```
+Client/UI  ──→  VMM (GetAppEnvEncryptPubKey)  ──→  KMS (GetAppEnvEncryptPubKey)
+                       pass-through proxy              actual key derivation
+```
+
+**Request**:
+
+```protobuf
+message AppId {
+  bytes app_id = 1;    // 20-byte app_id
+}
+```
+
+**Response**:
+
+```protobuf
+message PublicKeyResponse {
+  bytes public_key = 1;     // 32-byte X25519 public key
+  bytes signature = 2;      // Legacy k256 signature (no timestamp)
+  uint64 timestamp = 3;     // Unix timestamp in seconds when response was generated
+  bytes signature_v1 = 4;   // New k256 signature (with timestamp, replay-resistant)
+}
+```
+
+**HTTP call example** (prpc protocol):
+
+```
+POST {vmm_url}/prpc/Vmm.GetAppEnvEncryptPubKey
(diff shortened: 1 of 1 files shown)


kvinwang (commented):
## Review: Missing Security Caveat — Encrypted Env Lacks Origin Authentication

The spec accurately documents the encryption/decryption protocol, but as a **technical specification** it has an obligation to clearly state the security boundary of this scheme. Currently it only describes confidentiality, without warning that **there is no origin authentication**.

### The Problem

The encryption public key is derived from `app_id`, which is a public value (`SHA256(app-compose.json)[0..20]`). Anyone who knows the `app_id` can:

1. Call `VMM.GetAppEnvEncryptPubKey({ app_id })` to get the public key
2. Encrypt arbitrary env vars with it
3. Submit the encrypted payload to the VMM

The CVM will happily decrypt and use these forged env vars. The ECIES scheme guarantees **confidentiality** (only the TEE can decrypt), but provides **zero authenticity** — the CVM cannot distinguish between env vars from the legitimate developer and env vars from a malicious host operator or other attacker.

### What's Already Documented Elsewhere

This limitation is already acknowledged in:
- `docs/security/security-model.md` §"Environment variables need application-layer authentication"
- `docs/security/security-best-practices.md` §"Authenticated envs and user_config"

But a standalone spec document should not rely on readers having seen those pages.

### Suggested Addition

Add a **Security Considerations** section (before or after "End-to-End Flow"):

```markdown
## Security Considerations

### Encryption provides confidentiality, not origin authentication

The encrypted env scheme guarantees that only the target CVM (via its TEE-held private
key) can decrypt the environment variables. However, it does **not** authenticate who
encrypted them. Since the encryption public key is derivable from the public `app_id`,
**anyone** can encrypt and submit env vars to the CVM.

This means a malicious host operator (or any party with VMM access) could replace
the encrypted env payload with their own. The CVM will decrypt and use the replacement
without detecting the substitution.

### Developer responsibility: application-layer authentication

Developers **must** implement origin verification for encrypted env vars. Recommended
approaches:

1. **`APP_LAUNCH_TOKEN` pattern** (built-in support): Include a secret token as an
   encrypted env var. Its SHA256 hash is embedded in `app-compose.json`
   (`launch_token_hash` field), which is measured into the TDX attestation. A prelaunch
   script verifies the token hash before the app starts. The VMM UI auto-generates
   this prelaunch script — developers only need to add the `APP_LAUNCH_TOKEN` env var.
   See [security-best-practices.md](./security/security-best-practices.md#authenticated-envs-and-user_config).

2. **Custom signing**: Sign the env var payload with a key known only to the developer,
   and verify the signature inside the app at startup.

3. **Embedded secret**: Include a shared secret (known only to the developer and the
   app) in the encrypted env vars, and verify its presence at runtime.

For full guidance, see [security-best-practices.md](./security/security-best-practices.md)
and [security-model.md](./security/security-model.md#environment-variables-need-application-layer-authentication).
```

### Other Observations

1. **Spec accuracy**: The protocol details (X25519 + AES-256-GCM, no KDF, ciphertext format, key derivation chain) are consistent with the source code in `dstack-util/src/crypto.rs` and `kms/src/main_service.rs`. The protobuf definitions including `signature_v1` also match.

2. **Minor: HKDF salt not mentioned**: The key derivation uses HKDF-SHA256 with salt `b"RATLS"` internally. This is an implementation detail but could be noted for completeness.

3. **`user_config` parallel**: The best practices doc also notes that `user_config` has the same authenticity problem. Consider mentioning this in the spec if `user_config` interacts with encrypted env in any way.

PR #507, opened 2026-02-13
feat(kms): add auto-onboard support via config

Leechael:
## Summary

- Add `auto_onboard_url` config field to `OnboardConfig` for automating KMS onboarding from an existing instance
- When `auto_onboard_url` is set, the KMS automatically fetches keys from the source on startup, bypassing the Web UI
- On failure the process exits, relying on docker `restart: unless-stopped` to retry
- URL normalization (trim trailing `/`, ensure `/prpc` suffix) matches the existing Web UI onboard handler

## Test plan
- [x] `cargo check -p dstack-kms` passes
- [x] `cargo clippy -p dstack-kms -- -D warnings --allow unused_variables` passes
- [ ] Deploy two KMS instances, set `auto_onboard_url` on the second pointing to the first, verify auto-onboard completes

### DIFF
--- a/kms/src/onboard_service.rs
+++ b/kms/src/onboard_service.rs
@@ -337,7 +337,11 @@ pub(crate) async fn update_certs(cfg: &KmsConfig) -> Result<()> {
 }
 
 pub(crate) async fn auto_onboard_keys(cfg: &KmsConfig) -> Result<()> {
-    let source_url = cfg.onboard.auto_onboard_url.trim_end_matches('/').to_string();
+    let source_url = cfg
+        .onboard
+        .auto_onboard_url
+        .trim_end_matches('/')
+        .to_string();
     let source_url = if source_url.ends_with("/prpc") {
         source_url
     } else {
--- a/kms/src/onboard_service.rs
+++ b/kms/src/onboard_service.rs
@@ -159,12 +159,12 @@ struct Keys {
 }
 
 impl Keys {
-    async fn generate(domain: &str) -> Result<Self> {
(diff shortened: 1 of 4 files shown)


kvinwang (approved):
LGTM

kms/src/onboard_service.rs
```diff
@@ -327,6 +327,29 @@ pub(crate) async fn update_certs(cfg: &KmsConfig) -> Result<()> {
...
+        .trim_end_matches('/')
+        .to_string();
+    let source_url = if source_url.ends_with("/prpc") {
+        source_url
+    } else {
+        format!("{source_url}/prpc")
+    };
+    let keys = Keys::onboard(
+        &source_url,
+        &cfg.onboard.auto_bootstrap_domain,
```
kvinwang: `auto_bootstrap_domain` is optional in config (default is empty), but here it is passed directly as the target domain for `Keys::onboard`. Could we validate it first (e.g. reject empty/whitespace-only value with a clear error)? Otherwise auto-onboard may generate an unusable cert/domain setup when only `auto_onboard_url` is set.

— gpt-5.3-codex

Leechael: good catch, fixed in 45fcf7c8.

changes:
- add validate_domain() to reject empty/whitespace-only domains with clear error
- validate core.onboard.auto_bootstrap_domain in auto_onboard_keys() and bootstrap_keys() before key generation/onboarding
- also validate domain in update_certs() (both configured value and stored rpc_domain fallback)

cargo check -p dstack-kms passed.

PR #538, opened 2026-03-12
feat: on-chain TCB policy with dcap-qvl two-phase verification

kvinwang:
## Summary

- Upgrade `dcap-qvl` to `policy` branch with two-phase verification API (`QuoteVerificationResult` → policy validation → `VerifiedReport`)
- Propagate `QuoteVerificationResult` up through `DstackVerifiedReport` so KMS/verifier can apply app-specific TCB policies
- Add `IAppTcbPolicy` interface: both `DstackApp.sol` and `DstackKms.sol` store a versioned JSON TCB policy string on-chain
- `auth-eth` middleware reads policy from contract (with try-catch for backward compatibility with old contracts) and attaches it to `BootResponse`
- KMS parses the policy JSON (`{"version": 1, "intel_qal": [...]}`) and validates QVR against `RegoPolicySet`
- **Fail-close**: unknown policy versions are rejected
- Baseline policy rejects high-risk advisories: INTEL-SA-01397, INTEL-SA-01367, INTEL-SA-01314, INTEL-SA-00837

## Changes across layers

| Layer | Files | What |
|-------|-------|------|
| Solidity | `IAppTcbPolicy.sol`, `DstackApp.sol`, `DstackKms.sol` | On-chain policy storage (ERC-165 detectable) |
| TypeScript | `ethereum.ts`, `server.ts`, `types.ts` | Read policy from contract, pass in response |
| Rust (attest) | `attestation.rs`, `host_api.rs`, `verification.rs` | QVR propagation, baseline policy, advisory blacklist |
| Rust (kms) | `main_service.rs`, `upgrade_authority.rs` | Parse on-chain policy, RegoPolicySet validation |

## Depends on

- https://github.com/Phala-Network/dcap-qvl/pull/112 (will switch to crates.io version once released)

## Test plan

- [ ] `cargo check` passes (verified)
- [ ] `hardhat compile` passes (verified)
- [ ] `tsc --noEmit` passes (verified)
- [ ] Integration test with contract deploying TCB policy and KMS validating against it
- [ ] Backward compatibility: KMS works with old contracts that don't implement `IAppTcbPolicy`
- [ ] Fail-close: KMS rejects boot when policy version is unknown


dstack-attest/src/attestation.rs
```diff
@@ -195,20 +196,46 @@ impl QuoteContentType<'_> {
...
+                Ok(Some(qvr.validate(policy)?))
+            }
+            _ => Ok(None),
+        }
+    }
+
+    /// Get a [`TdxVerifiedReport`] without additional policy checks.
+    ///
+    /// Safe to call because the baseline policy was already validated during verification.
+    pub fn tdx_report(&self) -> Option<TdxVerifiedReport> {
```
kvinwang: Better remove this function.

dstack-attest/src/attestation.rs
```diff
@@ -375,7 +402,9 @@ impl GetDeviceId for () {
 impl GetDeviceId for DstackVerifiedReport {
     fn get_devide_id(&self) -> Vec<u8> {
         match self {
-            DstackVerifiedReport::DstackTdx(tdx_report) => tdx_report.ppid.to_vec(),
+            DstackVerifiedReport::DstackTdx(qvr) => {
+                qvr.clone().into_report_unchecked().ppid
```
kvinwang: More lightweight way?

kms/src/main_service.rs
```diff
@@ -204,6 +251,11 @@ impl RpcHandler {
         if !response.is_allowed {
             bail!("Boot denied: {}", response.reason);
         }
+        // Apply on-chain TCB policy (if set) to the TDX QuoteVerificationResult
+        if let Some(qvr) = att.report.tdx_qvr() {
+            validate_onchain_tcb_policy(qvr, &response.tcb_policy)
```
kvinwang: Better move this before the traditional is_app_allowed verification.
I'am not a fan of the is_app_allowed returning the policy. we should add a new api in auth_api.get_app_policy. As well for kms, auth_api.get_kms_policy.

PR #653, opened 2026-04-21
feature: TLS passthrough will select known app_id from TXT record w multiple entries

wwwehr:
## Problem

 As a phala cloud user/operator, I want to have two dstack-ingress instances share a
 given `_dstack-app-address.xxx` so that I can properly route a shared hostname
 using a load balancer. Currently, the router only accepts first dns record it 
 discovers which, also, isn't deterministic.

  ## Fix

  Extract `select_app_address` to iterate all values in the TXT record and return
  the first whose `app_id` is known to the local gateway's `apps` map. Falls back
  to the first parseable entry when none are locally known (preserves behavior
  for single-instance setups).

  The compat path (`_tapp-address` legacy) is unchanged.

  ## Tests

  - `test_resolve_app_address` — original live DNS integration test with mock state
  - `test_select_app_address_prefers_local` — multi-record, picks local app_id
  - `test_select_app_address_fallback_when_none_local` — multi-record, fallback to first

### DIFF
--- a/gateway/src/proxy/tls_passthough.rs
+++ b/gateway/src/proxy/tls_passthough.rs
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+use std::collections::{BTreeMap, BTreeSet};
 use std::fmt::Debug;
 use std::sync::atomic::Ordering;
 
@@ -39,8 +40,24 @@ impl AppAddress {
     }
 }
 
+/// find app by app id in current memory box
+fn select_app_address(items: &[Box<[u8]>], known_apps: &BTreeMap<String, BTreeSet<String>>) -> Result<AppAddress> {
+    let mut fallback = None;
+    for data in items {
+        if let Ok(addr) = AppAddress::parse(data) {
+            if known_apps.contains_key(&addr.app_id) {
+                return Ok(addr);
+            }
+            if fallback.is_none() {
+                fallback = Some(addr);
+            }
+        }
+    }
+    fallback.context("no app address found in txt record")
+}
+
 /// resolve app address by sni
-async fn resolve_app_address(prefix: &str, sni: &str, compat: bool) -> Result<AppAddress> {
+async fn resolve_app_address(prefix: &str, sni: &str, compat: bool, state: &Proxy) -> Result<AppAddress> {
     let txt_domain = format!("{prefix}.{sni}");
     let resolver = hickory_resolver::AsyncResolver::tokio_from_system_conf()
         .context("failed to create dns resolver")?;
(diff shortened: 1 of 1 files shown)


Leechael:
Thanks for putting this together. Before landing, two concerns about the DNS layer this operates on — the fix targets a layer that won't see the scenario it's meant to solve.

**1. RRset vs. character-strings within one record**

DNS has two independent "multi-value" axes at a TXT name:
- Multiple **records** in the RRset (each RR is a separate entry). Independent writers — e.g., two `dstack-ingress` instances each calling their provider's `set_txt_record` — produce this shape.
- Multiple `<character-string>` **within a single RR** (RFC 1035 §3.3.14). Normally used to encode a single logical value >255 bytes (SPF/DKIM), not alternatives. Producing this shape requires one coordinated writer.

In `hickory-resolver 0.24.4`, `TxtLookup::iter()` walks the RRset (yields `&TXT`, one per RR), and `TXT::txt_data()` returns the character-strings of *one* RR. The new code still does:

```rust
if let Some(txt_record) = lookup.iter().next() {     // picks ONE RR from the RRset
    if let Ok(addr) = select_app_address(txt_record.txt_data(), ...) {
        //                               ^^^^^^^^^^^^^^^^^^^^^^
        //                               iterates strings within that one RR
```

So `select_app_address` never sees a second writer's TXT in the multi-writer scenario — `lookup.iter().next()` already collapsed the RRset to one RR, nondeterministically. The tests pass because they feed `select_app_address` a `Vec<Box<[u8]>>` directly, simulating the single-RR-with-two-strings shape, which bypasses the `lookup.iter()` step. Against live DNS (where each writer produces one RR with one string), this code's output equals the old `.first()` path. The same structure exists in the `compat` branch and the wildcard-fallback branch.

If the intent is "iterate across the RRset", the shape is more like:

```rust
for txt_record in lookup.iter() {
    if let Some(data) = txt_record.txt_data().first() {
        // apply known-app selection against `data`
    }
}
```

**2. Upstream writer makes this scenario non-stationary regardless**

Even with the gateway-side read fixed, the reference ingress writer (`dstack-examples/custom-domain/dstack-ingress/scripts/dns_providers/base.py`) defines `set_txt_record`, `set_caa_record`, and `set_cname_record` as delete-then-create against any existing record whose content doesn't match the writer's own. Two instances with different `APP_ID:PORT` or different LE `accounturi` overwrite each other on every container start — the RRset has one element at a time, not two. CAA is strictly worse: RFC 8659 permits additive `issue` records, but the code deletes on tag match, so two LE accounts can't coexist without manual DNS editing.

As you note, the co-tenant scenario is reachable today only by manually pruning CAA and manually forcing two projects onto the same name — not a supported setup. That's fine as a niche, but it means the premise for "TXT records will contain multiple app_id candidates the gateway should choose from" depends on out-of-band manual DNS edits that also have to survive every ingress restart.

**Suggestion**

If the goal is to keep the door open for externally-coordinated multi-writer setups:
1. Move the iteration to `lookup.iter()` so the gateway actually reads the RRset rather than one RR's strings (apply in all three call sites — primary, compat, wildcard).
2. Document the preconditions (manual DNS co-management, additive CAA, TTL vs. restart-rewrite races) so operators don't expect this to work with stock `dstack-ingress`.

If the goal is just to make TXT selection deterministic on the common single-record path, the smaller change is fine — but then the PR description should drop the "two shared ingress instances" framing, since the upstream writer prevents that RRset shape from appearing on the wire.

Happy to iterate on this.

PR #656, opened 2026-04-22
fix: prevent u64 underflow panic in online instance count

bravesasha:
Bare subtraction now - ts panics in debug builds and silently wraps around in release builds when a peer reports a future timestamp due to clock skew. Use saturating_sub to handle this safely.


h4x3rotab:
Thanks for the contribution!

PR #673, opened 2026-05-11
Fix ra-tls certificate SAN generation for IP-only endpoints

ameba23:
`ra_tls::cert::CertRequest::into_cert_params` converts every configured subject alternative name into `SanType::DnsName`, so an IP literal like `203.0.113.10` is encoded as a DNS SAN.

https://github.com/Dstack-TEE/dstack/blob/4b7372db3ec338a9262df6664e5485ec70fd33a6/ra-tls/src/cert.rs#L372-L378

rustls/webpki requires IP endpoints to be represented as actual IP SANs, so name verification fails when clients connect by IP address.  This makes it difficult to use `ra_tls` for IP-only hosts.  See related issue in our attested-tls implementation which depends on `ra-tls`: https://github.com/flashbots/attested-tls/issues/30

This PR delegates SAN parsing to `rcgen::CertificateParams::new(...)`, which internally converts IP literals into `SanType::IpAddress` and leaves hostnames as DNS SANs.

So this **is** a behavioral change, but it only effects names which successfully parse as IP addresses.

A test covering mixed SANs is included which checks that:
- `203.0.113.10` is encoded as `SanType::IpAddress`
- `test.example.com` remains `SanType::DnsName`

### DIFF
--- a/ra-tls/src/cert.rs
+++ b/ra-tls/src/cert.rs
@@ -11,7 +11,7 @@ use anyhow::{anyhow, bail, Context, Result};
 use fs_err as fs;
 use rcgen::{
     BasicConstraints, Certificate, CertificateParams, CustomExtension, DistinguishedName, DnType,
-    ExtendedKeyUsagePurpose, IsCa, KeyPair, KeyUsagePurpose, PublicKeyData, SanType,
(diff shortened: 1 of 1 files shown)


kvinwang (approved):
LGTM, thanks for the fix!

PR #674, opened 2026-05-11
gw: add admin API token authentication

kvinwang:
## Summary

The gateway admin server has historically had **no application-layer auth** — `AdminRpcHandler::construct` doesn't inspect the caller at all, so anything reachable on the admin port (default `0.0.0.0:8001` inside `dstack-app/builder/entrypoint.sh`) can hit any admin RPC or load the dashboard. Combined with the dashboard's view of cluster state and the fact that DNS-credential redaction was the only recent protection layer, this is an obvious next hardening step.

This PR adds an opt-in shared-secret check, modelled on KMS `ensure_admin` but storing the **plain token** in config (no SHA-256 hash) so it can be injected directly via dstack **encrypted env** without an extra hashing step on the operator side.

### Changes

- `AdminConfig.admin_token` (empty = no auth, with a startup `WARN` so existing deployments keep their current behaviour).
- New `gateway/src/admin_auth.rs` — a Rocket `Fairing` attached to the admin instance. Accepts the token via:
  - `X-Admin-Token: <token>` header (any HTTP method)
  - `Authorization: Bearer <token>` header (any HTTP method)
  - `?token=<token>` query parameter **GET only** (intended for dashboard browser links). After a successful auth the `token` query param is **stripped from the request URI** so it does not propagate to access logs, downstream handlers, or the Referer header.

  Constant-time compare via `subtle::ConstantTimeEq`. Rejected requests are rewritten to `/__admin_unauthorized`, a sentinel route with handlers for **GET / POST / PUT / PATCH / DELETE / OPTIONS / HEAD** so every method returns 401 (not 404/405).
- Tests: `gateway/src/admin_auth.rs` includes a `mod tests` covering token-extraction precedence (header > Bearer > query), the GET-only restriction for query token, query-token stripping, wrong-token rejection, empty-token (auth disabled), and 401 across all HTTP methods. Uses Rocket's `local::asynchronous::Client` — no port binding needed.
- `dstack-app` plumbing: `ADMIN_TOKEN` env in `docker-compose.yaml`, written into `[core.admin] admin_token` by `builder/entrypoint.sh` (with the existing `validate_env` toml-injection guard).
- `bootstrap-cluster.sh` now sends `X-Admin-Token` when `ADMIN_TOKEN` is set, so cluster bootstrap continues to work after enabling auth.
- `subtle` declared in `[workspace.dependencies]` for consistency with the rest of the workspace.

### Why a fairing + URI rewrite (not per-route guards)

The admin server mounts `prpc!(...)` macro-generated routes plus the dashboard route. Adding request guards to those would require modifying the macro. A `Kind::Request` fairing that rewrites the URI to a 401 sentinel covers every current and future route on the admin instance with one attachment, and is the canonical Rocket pattern for early rejection in a fairing.

### Backward compatibility

- Default `admin_token = ""` keeps existing deployments unchanged; the startup log emits `WARN admin server enabled without admin_token; admin API is exposed without authentication`.
- Once an operator sets `ADMIN_TOKEN`, all admin endpoints require it (including the dashboard — visit `https://gateway/?token=<token>` once, or send the header).

## Test plan

- [x] `cargo check -p dstack-gateway` clean
- [x] `cargo clippy -p dstack-gateway -- -D warnings -D clippy::expect_used -D clippy::unwrap_used --allow unused_variables` clean (CI flags)
- [x] `cargo fmt --check --all` clean
- [x] `cargo test -p dstack-gateway admin_auth` — 10 tests pass
- [ ] Manual: start gateway with `admin_token = "s3cr3t"` and verify:
  - `curl http://127.0.0.1:8011/prpc/Status` → 401
  - `curl -H "X-Admin-Token: s3cr3t" http://127.0.0.1:8011/prpc/Status` → 200
  - `curl -H "X-Admin-Token: wrong" http://127.0.0.1:8011/prpc/Status` → 401
  - `curl -H "Authorization: Bearer s3cr3t" http://127.0.0.1:8011/prpc/Status` → 200
  - `curl "http://127.0.0.1:8011/?token=s3cr3t"` → 200 dashboard (URI stripped)
  - `curl -X POST "http://127.0.0.1:8011/prpc/Status?token=s3cr3t"` → 401 (query token rejected on POST)
  - `curl http://127.0.0.1:8011/` → 401
  - `curl -X OPTIONS http://127.0.0.1:8011/prpc/Status` → 401 (not 404/405)
- [ ] Manual: start gateway with `admin_token = ""` and confirm WARN log + open access (legacy behaviour)
- [ ] Manual: e2e cluster bootstrap via `bootstrap-cluster.sh` with and without `ADMIN_TOKEN`

## Addressed review feedback (Copilot)

| # | Comment | Fix |
| --- | --- | --- |
| 1 | `?token=` query param risks leaking via access logs / history / Referer | Restricted query token to **GET only**; URI is stripped of `token` param after successful auth |
| 2 | Sentinel only had GET/POST handlers — other methods would 404/405 | Added handlers for PUT / PATCH / DELETE / OPTIONS / HEAD |
| 3 | No tests for the new auth path | Added `mod tests` with Rocket local client — 10 cases covering all extraction paths + all-method 401 |
| 4 | Startup log mentioned only `X-Admin-Token` | Simplified to "admin server authentication enabled" |
| 5 | `subtle` pinned locally instead of via workspace | Moved to `[workspace.dependencies]` and `subtle.workspace = true` |


Leechael (approved):
lgtm

PR #675, opened 2026-05-11
Protect gateway admin API with token auth

Leechael:
## Summary
- require a Bearer token for the gateway admin dashboard and Admin pRPC routes
- add admin token config with env fallbacks and an explicit insecure test escape hatch
- wire deployment/bootstrap scripts and docs to pass the admin token

## Verification
- prek run --from-ref HEAD^ --to-ref HEAD --show-diff-on-failure
- cargo fmt --check -p dstack-gateway
- cargo check -p dstack-gateway
- cargo test -p dstack-gateway admin_auth
- git diff --check


Leechael:
Closing this draft in favor of #674. The fairing-based approach in #674 is broader, better tested, and already green in CI. Any remaining dashboard auth UX discussion should continue there.

PR #688, opened 2026-05-19
chore(ci): bump builder pinned packages for Debian upstream drift

Leechael:
Fix CI failures in gateway and kms builder image checks caused by Debian package version drift.

Updated packages:
- `libssl-dev` / `libssl3t64`: 3.5.5-1~deb13u1 → 3.5.6-1~deb13u1
- `musl` / `musl-dev` / `musl-tools`: 1.2.5-3 → 1.2.5-3.1~deb13u1
- `openssl` / `openssl-provider-legacy`: 3.5.5-1~deb13u1 → 3.5.6-1~deb13u1


kvinwang:
Pushed two follow-up commits because the original version bumps only paper over a bug in `pin-packages.sh` — the snapshot pinning was never actually in effect on the trixie-based builders.

**Root cause.** `pin-packages.sh` writes `/etc/apt/sources.list` (legacy format) pointing at the snapshot, hard-coded to `bookworm`. That works for the verifier image (`rust:1.92.0-bookworm`, which uses the legacy sources file). But `kms` and `gateway` use `rust:1.92.0`, which is now based on **Debian 13 (trixie)** and ships its sources via **deb822** at `/etc/apt/sources.list.d/debian.sources`. That file was left untouched, so apt kept resolving from live `deb.debian.org/debian trixie` instead of the snapshot. The packages drift every time upstream publishes a security update — which is exactly what we saw (libssl 3.5.5 → 3.5.6, musl 1.2.5-3 → 1.2.5-3.1~deb13u1). Bumping the pinned versions would just kick the can down the road until the next trixie point release.

**Fix.** `pin-packages.sh` now:
1. Detects the suite from `/etc/os-release` (`VERSION_CODENAME`), so the same script works for trixie and bookworm without hard-coding.
2. Wipes **both** `/etc/apt/sources.list` and `/etc/apt/sources.list.d/*.{list,sources}` before writing the snapshot sources, so the base image's default live sources can no longer leak through.

Verified locally inside both `rust:1.92.0` (trixie) and `debian:bookworm`: `apt-get update` only contacts `snapshot.debian.org`, and `apt-cache madison libssl3t64` resolves to `3.5.5-1~deb13u1` from snapshot trixie/main — the originally baselined version. So the version bumps in the first commit have been reverted; the pinned files now match what the snapshot actually serves.

The third commit just silences a shellcheck SC1091 on the `. /etc/os-release` sourcing that the `prek` check flagged.

Issue #699, opened 2026-05-28
Onboarding audit: cut 'time to first dstack app' from ~22 steps to a single script


h4x3rotab:
## The problem

If you have a TDX host and want to run your first dstack app, the README says "deploy your own" and links to the deployment guide. Following that guide end-to-end takes around **22 ordered steps across two repos** before a `docker-compose.yaml` is reachable in a browser.

The first time I walked through it I had:

- Three terminals open (vmm, auth-simple, log tail)
- One browser tab on the KMS bootstrap page
- One text editor on `auth-config.json` that I edited four separate times
- A Cloudflare API token I had to provision because the gateway needs DNS-01 ACME
- A domain I had to own because the gateway URL pattern is `<id>-<port>.gateway.<your-domain>`

We can do better. This issue is about figuring out how.

## What's painful, concretely

A few moments where I sat there and thought "this should not be a step":

**The chicken-and-egg dance with the KMS allowlist.** On a fresh `auth-config.json`, `kms.mrAggregated` is empty, and `auth-simple` denies all KMS boots against an empty allowlist — but you can't know the value until the KMS CVM has booted. So the documented procedure is: deploy the KMS, watch bootstrap fail, call `Onboard.GetAttestationInfo` to read the hash, paste it into `auth-config.json`, retry. For a single-node KMS this gate is near-circular — it's the operator's KMS proving its own measurement to an allowlist the same operator just wrote — and removing it is a config default, not a hard problem (see the KMS section below).

**The "exit 1, edit the .env, run me again" pattern.** Three different scripts do this: `kms/dstack-app/deploy-simple.sh`, `gateway/dstack-app/deploy-to-vmm.sh`, `gateway/dstack-app/bootstrap-cluster.sh`. On the happy path that's three re-invocations just to get past prompts that could have been CLI flags or interactive prompts.

**Copy-paste a hash, then type `y`.** When you deploy the gateway, the script prints a compose hash, asks `Continue? [y/N]`, and the right answer is *not yet* — you're supposed to put the hash in `auth-config.json` first, then come back and type `y`. Same dance again when you deploy your first app. The script knows the hash. The auth file is on the same disk. It could just write it.

**Three terminals, no systemd.** vmm, auth-simple, and the log tail all run as foreground processes. There's no installed service, no `systemctl status`, no `journalctl`. Restarts after config edits are manual. Lifecycle is the operator's problem.

[...]

PR #731, opened 2026-06-18
feat(cli): dstack + dstackup — hands-off single-node onboarding

h4x3rotab:
Getting a first app onto dstack today takes around 20 manual steps — and you have to know which guest-image version to download. This adds a single-node onboarding path that's a few commands and figures the image out for you:

```bash
# fetch the latest guest image (cpu; --gpu for nvidia), checksum-verified
dstackup image pull

# set up the host + bootstrap a KMS-in-CVM (auto-detects TDX vs SEV-SNP, picks the image)
dstackup install

# deploy an app
dstack deploy <compose> --port <p> --allowlist <path>
```

Two binaries, modeled on kubeadm/kubectl: `dstackup` for host setup (privileged, local) and `dstack` for deploys (local socket or remote http). The dashboard binds localhost and is reached over an SSH tunnel, which also gives the browser a secure context for the env-encryption crypto without needing a cert.

Four new crates under `crates/`; no existing crate is touched:

- `dstack-cli-core` — shared lib: VMM prpc client, config rendering (`vmm.toml`/`kms.toml`/`auth-allowlist.json`), app-compose build, host/SGX detection, CID/port coexistence scan, atomic-write + flock helpers.
- `dstack-cli` (binary `dstack`) — `deploy`/`apps`/`logs`, global `-j/--json`; command names follow the `phala` CLI.
- `dstackup` — `install`/`status`/`image`/`destroy`. Manages `dstack-vmm` + `dstack-auth` as systemd units and bootstraps the single-node KMS-in-CVM; idempotent, with deterministic teardown.
- `dstack-auth` — Rust rewrite of the single-operator KMS auth webhook (compose-hash + os-image allowlist, fails closed).

Part of #699, where the design rationale and roadmap live.

## Guest images

`dstackup image pull [--version <v>] [--gpu]` fetches a guest image from meta-dstack releases — one image serves both TDX and SEV-SNP, so the only choice is cpu vs gpu — and `image list`/`image rm` manage what's local. The download is streamed and sha256-verified against the release's published digest (fails closed; `--insecure` to override when a release publishes none), then staged and atomically adopted so a truncated download or a half-unpack can't masquerade as a valid image. `install` auto-selects the newest local image when `--image` is omitted, and points you at `image pull` when none is present. Images live in `<prefix>/images`, resolved identically for `install` and `image` so they can't drift.

## Single-node trade-offs

The flow makes a few scoped relaxations. Each is confined to the single-node KMS-in-CVM config and none touch the per-app key path or multi-node replication:

- `enforce_self_authorization=false` drops only the KMS self-attestation gate at bootstrap; per-app quote verification and the compose-hash allowlist stay enforced.
- `verify_os_image=false` is compensated by pinning the app's measured OS image (`digest.txt`) into the webhook allowlist, and it's fail-closed: a missing pin aborts `install` unless `--allow-unpinned-image` is passed.
- `--expose` is disabled. The VMM management RPCs are unauthenticated and the TLS+token transport isn't built yet, so it fails with SSH-tunnel guidance instead of opening an unauthenticated control plane.
- TCB-status enforcement is deferred (the operator trusts their own host, and real TDX hosts routinely report a non-UpToDate TCB) — noted in `dstack-auth`.

Allowlist updates are atomic (temp + rename + dir fsync) under an flock, and `install` runs all CID/port collision checks before any side effect, so a clash refuses cleanly instead of half-installing.

## Validation

End-to-end on real hardware:

- Intel TDX: `dstackup install` (dstack-vmm built from this repo + meta-dstack v0.5.11) → KMS-in-CVM bootstraps → `dstack deploy nginx` → HTTP 200 with the os-image pin enforced → clean `destroy`. Fail-closed paths (missing pin, port collision) bail with no side effects; CID/host-api coexistence checked against a second running VMM.
- AMD SEV-SNP: platform auto-detected, SNP `vmm.toml`/`kms.toml` rendered, a from-source guest image boots (SEV active, VMPL0) and serves an app over HTTP 200. This surfaced and fixed a `secure_time` default that reboot-loops the guest, and filed #745 for the underlying chrony-NTS image bug.
- `image pull` exercised against live meta-dstack releases: sha256 verified, atomically adopted, re-pull is a no-op.

`cargo fmt`, `cargo clippy -- -D warnings -D clippy::expect_used -D clippy::unwrap_used`, and the unit tests pass.

## Not in this PR (tracked in #699)

Remote transport (TLS+token for `--expose`, remote `dstack logs`, `--token`), the gateway tier, env-var encryption for `dstack deploy`, fuller phala-CLI alignment, and building/packaging the guest OS image itself.


docs/onboarding-redesign.md
h4x3rotab, line 1: We should separate the work progress tracking doc from the user facing docs

crates/dstack-cli/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+    token: Option<String>,
+
+    #[command(subcommand)]
+    command: Command,
+}
+
+#[derive(Subcommand)]
+enum Command {
+    /// Deploy an app from a docker-compose file.
+    Run {
```
h4x3rotab: let's try to make our cli align with `phala` cli as much as we can

h4x3rotab (commented):
- crates/dstack is confusing. Let's call it `crates/dstack-cli`, but the binary can still be called dstack
- crates/dstack-core is also confusing. because it looks like the core of the entire dstack project. but it's actually just a shared lib between the two clis
- there are so many mention of "2025 Phala Network". it's 2026 now

docs/onboarding-redesign.md
h4x3rotab: Done in 1661cc50 — removed `docs/onboarding-redesign.md`. It's a planning/roadmap artifact (design rationale, "what we validated", roadmap, open questions), now tracked only in #699, so `docs/` stays user-facing-only. The actual user-facing quickstart is a separate follow-up.

crates/dstack-cli/src/main.rs
h4x3rotab: Done in 1661cc50 — aligned the client with the `phala` CLI: `run` → `deploy`, `ls` → `apps`, dropped the separate `upgrade` (phala folds update into `deploy`), and added a global `-j/--json` (honored by `deploy`/`apps`). `logs` already matched; the binary is still `dstack`. Fuller alignment (`ps`, `--cvm-id`, a `phala.toml`-style project file) is noted as a follow-up in #699 so it can be designed deliberately.

h4x3rotab:
Thanks for the review — all five points are addressed:

1. `crates/dstack` → **`crates/dstack-cli`** (binary still `dstack`) — 9be329b6
2. `crates/dstack-core` → **`crates/dstack-cli-core`** (it's the shared lib for the two CLIs, not the project core) — 9be329b6
3. `© 2025` → **`© 2026`** across the new files — 9be329b6
4. Removed the planning/roadmap doc `docs/onboarding-redesign.md` (tracked in #699) — 1661cc50 ↑
5. Aligned the CLI with `phala` (`deploy`/`apps` + global `-j/--json`) — 1661cc50 ↑

Two extra cleanups while in here: de-jargoned "Tier-1" in code comments → "single-node (no gateway)" (47f4c475), and split the 925-line `dstackup/main.rs` into focused modules `cli`/`install`/`destroy`/`state`/`systemd`/`cid` (ac27bca0).

Verified locally against the exact CI flags (`clippy -D warnings -D clippy::expect_used -D clippy::unwrap_used`, fmt, tests); CI is re-running on the latest push.

PR #756, opened 2026-07-07
VMM: networking mode management

Leechael:
## Summary

Adds multi-NIC networking support to the VMM and UI:

- supports repeated per-VM network configurations while retaining legacy `networking` manifest compatibility;
- exposes network configuration and capabilities over VMM RPC;
- creates one QEMU netdev/device pair per interface with deterministic per-NIC MAC addresses;
- validates bridge/custom-mode requests and keeps runtime networking snapshots consistent;
- measures the NIC count in RTMR0 so multi-NIC CVMs remain attestable;
- adds create/update controls to the VMM UI.

Rebased and relocated onto the unified monorepo layout from PR #770.

## Validation

- `cargo test -p dstack-vmm -p dstack-mr -p dstack-types -p dstack-verifier`
- `npm ci && npm run build` in `dstack/vmm/ui`
- `reuse lint`
- `prek run --all-files`


kvinwang:
A note on the `guest_ips` addition before we grow it further:

The whole `guest_ip` machinery only exists to serve VMM's built-in port-forward feature. And the way it learns the guest IP is pretty awkward — it has to talk to an *external* DHCP server out-of-band (dnsmasq `--dhcp-script` → `dhcp-notify.sh` → `ReportDhcpLease` RPC) just to find out what address the guest got. So the stored value is really "what dnsmasq handed out," not the guest's actual IP: it goes stale (the `del`/release path is a no-op), it's empty for user/slirp NICs, and the RPC that feeds it is unauthenticated. That's a lot of moving parts for what it delivers.

Context worth flagging: the built-in port forwarding — and `guest_ip` along with it — was already a candidate for removal; the intent was to deprecate and delete it rather than build on top of it.

Given that, I'd suggest **not** extending this into a per-MAC `guest_ips` map / `guest-ips.json` file plus the new `interfaces[].ip` surface. That doubles down on a subsystem we meant to drop: it adds a second on-disk file with the same staleness/trust caveats and grows the public API contract around it. If multi-NIC status reporting is genuinely needed, let's settle the fate of port-forward first — and if it stays, the DHCP-report path should get authentication + a release/cleanup path before we persist and expose it more widely.

Happy to help sketch the removal path if we decide to go that way.

kvinwang (approved):
LGTM

kvinwang:
Rebased onto current `master` (`7b1b4934`, including #787) and force-pushed.

The former `qemu.rs` changes were reapplied by responsibility rather than restoring the monolith:

- network resolution/validation and indexed MAC derivation → `app/network.rs`
- RPC networking presentation and interface status → `app/vm_info.rs`
- manifest compatibility and runtime-network snapshots → `app/workdir.rs`
- repeated `-netdev`/`virtio-net-pci` construction → the prepared `QemuCommandBuilder` path in `app/qemu.rs`

I also kept bridge/filesystem validation in `PreparedQemuLaunch`, so command construction remains host-I/O-free, and added coverage for two NICs plus first-user-NIC-only host forwarding. Range-diff confirms original commits 2–9 replay unchanged; the first commit differs only where its monolithic QEMU changes were relocated.

Validation:

- `cargo test -p dstack-vmm -p dstack-mr -p dstack-types -p dstack-verifier`
- `cargo clippy -p dstack-vmm --all-targets --all-features -- -D warnings`
- `npm ci && npm run build` in `dstack/vmm/ui`
- `prek run --all-files`
- `reuse lint` on a clean archive
- `./run-tests.sh`

PR #818, opened 2026-07-23
fix(simulator): mount host share before startup

kvinwang:
## Summary

- add `dstack-util host-shared mount|unmount` and reuse it from guest setup
- mount the host share before `dstack-tee-simulator` starts, then unmount it after the simulator is ready
- move development-only simulator settings out of `.sys-config.json` into `.tee-simulator.json`
- make VMM emit the separate simulator config, including the matching mr/vm config documents
- update mock-attestation, simulator E2E, and no-TEE development documentation
- fix the dev-image simulator recipe for built-in FUSE and aws-lc Yocto cross builds

## Root cause

`dstack-tee-simulator.service` runs before `dstack-prepare.service`, but the simulator previously tried to read `/dstack/.host-shared/.sys-config.json`. The host share is mounted and copied by prepare, so it was not visible yet. The simulator then started without its required mock seed, prepare failed through its dependency, and the guest entered a reboot loop.

The new early mount is simulator-specific at the systemd layer, while the mount implementation is a shared `dstack-util` command also used by production setup. Simulator-only settings no longer alter the production sys-config schema.

## Validation

- `cargo check --manifest-path dstack/Cargo.toml -p dstack-util -p dstack-vmm -p dstack-tee-simulator -p mock-attestation`
- `cargo test --manifest-path dstack/Cargo.toml -p dstack-types -p dstack-util -p dstack-vmm -p dstack-tee-simulator -p mock-attestation`
- `dstack/tests/e2e/attestation/run.sh` (all six simulated platforms)
- `prek run --all-files`
- `git diff --check`

A clean Yocto dev-image build was also started and passed parsing plus 3,700+ of 5,502 prerequisite tasks before being stopped for time; full image boot/persistence validation remains for CI or follow-up testing.


kvinwang:
Addressed the review feedback in 717cc3abc:

- simulator startup is now gated only by the presence of host-provided `.tee-simulator.json`; TEE device/provider detection was removed
- when the file is absent, the condition unmounts the early host share and skips the simulator without reporting a configuration error
- restored disk/9p mount and fallback logs
- added a VMM unit test covering separate simulator config emission, mr/vm config injection, and stale-file removal when disabled
- cleaned the simulator fixture, clarified the standalone default config path, and fixed the new lowercase error contexts
- documented that the guest starts the simulator only when the VMM supplies the file

Validation:
- VMM/util/simulator tests pass (50 + 53 + simulator unit/process tests)
- all six attestation E2E platforms pass; NitroTPM had one transient vTPM device race and passed immediately on isolated retry
- `prek run --all-files` and `git diff --check` pass

Yocto retry reached 67% current sstate, then failed while linking the upstream sqlite3 target after the earlier interrupted/concurrent build left the build directory inconsistent. This is outside the changed recipes; full image boot validation is still outstanding.

kvinwang:
Follow-up in f2ce587af moves simulator selection to the VM instance:

- added optional `VmConfiguration.simulated_tee` RPC field (platform string)
- persisted the selection in each VM manifest and returned it through VM status/config
- added `dstack deploy --simulated-tee <platform>`; selecting it makes that instance non-TEE
- added the matching platform selector to the create/clone UI
- node-level `[cvm.tee_simulator]` now supplies only the seed/collateral template and no longer enables simulation or selects the emitted platform
- `.tee-simulator.json` is written only for manifests with `simulated_tee`; the instance platform overrides the node template
- one-shot mode follows the same per-instance behavior
- key-provider selection remains independent: TPM only controls whether application keys need a TPM; simulator enablement is never inferred from it. For simulated GCP TDX/NitroTPM, the simulator-provided TPM avoids starting a duplicate QEMU swtpm.

Validation:
- all 54 dstack-vmm tests pass, including per-instance persistence, invalid/missing-node-config rejection, template override, and TPM cross-product behavior
- VMM UI protobuf generation and production build pass
- Python CLI compiles
- `prek run --all-files` and `git diff --check` pass

kvinwang:
Follow-up in 7bf167e77 moves the QEMU swtpm decision to deployment and makes it part of attested VM configuration:

- deployment parses the app compose key provider and combines it with the instance's `simulated_tee`
- persists the derived `qemu_swtpm: Option<bool>` in the VM manifest; new manifests always store `Some(bool)`
- QEMU consumes the persisted switch; `None` retains a lazy fallback only for old manifests
- serializes enabled `qemu_swtpm` into guest `vm_config` (false is omitted for wire/cache compatibility)
- propagates the flag through dstack-mr and verifier `Machine` construction
- ACPI reconstruction attaches the same `tpm-tis` device shape; dstack-mr starts a temporary swtpm backend while invoking `dstack-acpi-tables`
- verifier/KMS images and external-build docs now include the swtpm runtime dependency

The deployment decision matrix is covered:
- key-provider TPM + no simulator / dstack TDX simulator => QEMU swtpm
- key-provider TPM + GCP TDX or NitroTPM simulator => simulator TPM, no QEMU swtpm
- non-TPM key provider => no QEMU swtpm

Validation:
- dstack-types: 35 passed
- dstack-mr: 27 passed
- verifier: 14 passed
- VMM: 56 passed
- prek and diff checks pass
- manual `dstack-acpi-tables` reconstruction with a live swtpm confirmed the no-TPM and tpm-tis ACPI blobs have different SHA-256 hashes:
  - no TPM: `b8ad577b2a15919e67320b1c148b1b3cf0fef77e457c0b458c6f37541d0b6d4a`
  - tpm-tis: `7a97782723b6ddc0fc7f5f8deae6230a28be9e0b04a8297109693440afa08677`

kvinwang:
Additional fix in c884dbbc9: VM updates now recompute and persist `qemu_swtpm` from the updated app-compose key provider plus the existing instance `simulated_tee`. This also migrates legacy manifests on update. The dstack-mr swtpm ACPI helper import was fixed; VMM/dstack-mr tests and prek pass.

PR #841, opened 2026-07-31
test: add colocated core component acceptance suite

kvinwang:
## Summary

This PR adds the core-component acceptance suite and organizes the complete suite directly under `test-suites/` so specifications, automation, fixtures, inventories, manifests, and retained audit evidence have one clear home.

- Adds the deterministic `dstack-test` runner, dashboard, registry validation, fixture providers, and component harnesses.
- Adds 358 acceptance cases covering guest OS, VMM, KMS, gateway, verifier, and cross-component integration behavior.
- Colocates single-case entrypoints with their `case.md` specifications as `run.py`, `run.sh`, or `run.cjs`.
- Keeps multi-case and location-sensitive harnesses in `shared/automation/`.
- Distributes metadata across suite, chapter, section, and case directories; the runner discovers cases in directory-name order.
- Uses each case-local `metadata.json` as the authoritative case-to-entrypoint binding.
- Keeps fixture providers and images under `shared/fixtures/`, inventories under `catalog/`, environment manifests under `manifests/`, and retained historical evidence under `audit/`.
- Adds suite authoring, execution, reporting, and methodology documentation.
- Excludes product/runtime changes and source-local unit tests; those changes were split into dedicated PRs.

## Layout

```text
test-suites/
├── runner/
├── cases/
├── shared/
│   ├── automation/
│   └── fixtures/
├── catalog/
├── manifests/
└── audit/
```

## Separation invariant

The PR targets `next`. Its changed paths are limited to:

- `REUSE.toml`
- `docs/testing/**`
- `test-suites/**`

There are no product/runtime or source-local test changes under `dstack/**` or `os/**`.

## Verification

- Registry: 357 scripted / 357 promoted / 0 findings (358 total cases)
- Runner unit tests: 37 passed
- `prek run --all-files`: passed
- `reuse lint`: passed
- `git diff --check origin/next...HEAD`: passed

## Rebase and tdxlab validation (2026-09-18)

- Rebases the branch onto `next` at `e2cf39ae01`. Final PR head: `118dba1692`.
- Adds and corrects coverage for the 82 changes merged since the previous baseline `afbafb90aa`. See `test-suites/audit/core-components-next-rebase-audit-2026-09-17.md` for the per-PR mapping.
- New cases: `tc-gos-guestapi-006` (GuestApi.GpuInfo), `tc-gos-setup-026` (`dstack-util gpu-info`), and `tc-vmm-install-007` (source installer). The suite now holds 361 cases, 360 of them scripted.
[...]

PR #845, opened 2026-07-31
fix(simulator): tolerate udev TPM node race

kvinwang:
This PR is withdrawn because its premise was incorrect.

The simulator mode already determines which device ABI is created in `main.rs`:

- `dstack-gcp-tdx` enters `start_gcp_vtpm`;
- `dstack-aws-nitro-tpm` enters `run_nitro_vtpm`;
- non-TPM simulator modes do not create a TPM device.

Both TPM paths already reject `/dev/tpm0` or `/dev/tpmrm0` if present before startup. Once the selected mode registers its vTPM, it strictly runs `mknod`; any collision, including a system-created node appearing first, is an error.

The proposed tolerance/configuration layer weakened that strict ownership rule and is not needed. No dedicated acceptance case justified silently adopting an existing node. The associated #841 test changes have been reverted.

### DIFF
--- a/dstack/dstack-types/src/lib.rs
+++ b/dstack/dstack-types/src/lib.rs
@@ -922,7 +922,7 @@ impl SysConfig {
     }
 }
 
-#[derive(Deserialize, Serialize, Debug, Clone, Default)]
+#[derive(Deserialize, Serialize, Debug, Clone)]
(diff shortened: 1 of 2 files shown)


kvinwang:
Closing as rejected: the original platform-selected strict creation behavior is the intended design.

PR #840, opened 2026-07-31
fix: resolve core component acceptance findings

kvinwang:
## Summary

This PR isolates the product-code fixes discovered while executing the core-component acceptance campaign.

- Contains product changes only under `dstack/**` and `os/**`.
- Fixes issues across Guest OS, KMS, Gateway, Verifier, VMM, supervisor, certificate handling, attestation, storage, and compatibility paths.
- Preserves newer upstream Gateway TLS termination behavior while applying the compatible fixes from the acceptance branch.
- Excludes the acceptance plan, test harnesses, fixtures, scripts, and result artifacts; those are in a separate stacked PR.

## Separation invariant

Mechanical comparison against `master` reports 129 changed files, all under `dstack/**` (118) or `os/**` (11).

## Verification

- `cargo fmt --all -- --check`: passed
- `cargo check --workspace --all-targets`: passed (warnings only)
- `git diff --check origin/master..HEAD`: passed

The full acceptance campaign reached 344 PASS and 8 hardware-only BLOCKED cases out of 352, with no FAIL, ERROR, or INCOMPLETE cases. Build-correctness cases were removed from the plan as requested.

PR #855, opened 2026-07-31
fix(guest): report ZFS pool creation errors

kvinwang:
## Problem

The ZFS setup path discarded the command failure that occurs while creating the application pool. Boot then continued and failed later against a missing pool, obscuring the actual `zpool create` error.

## Root cause and fix

Propagate the pool-creation command error immediately, including its diagnostic output, so provisioning stops at the real failure.

## Implementation

- `fix(guest): report ZFS pool creation errors`

Changed paths:

- `dstack/dstack-util/src/system_setup.rs`

## Scope

This PR addresses one logical `guest` finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

## Verification

- `git diff --check origin/master..origin/codex/fix-guest-zfs-pool-errors`: passed.

### DIFF
--- a/dstack/dstack-util/src/system_setup.rs
+++ b/dstack/dstack-util/src/system_setup.rs
@@ -2380,11 +2380,20 @@ impl<'a> Stage0<'a> {
             match opts.storage_fs {
                 FsType::Zfs => {
                     info!("Creating ZFS filesystem");
-                    cmd! {
-                        zpool create -o autoexpand=on dstack $fs_dev;
-                        zfs create -o mountpoint=$mount_point -o atime=off -o checksum=blake3 dstack/data;
+                    let output = Command::new("zpool")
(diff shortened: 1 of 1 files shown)


kvinwang:
Closing this because the original code already propagates `zpool create` failures.

`cmd!` is imported as `cmd_lib::run_fun`. For a command group, `GroupCmds::run_fun()` executes preceding commands through `self.run_cmd()?`, so a non-zero `zpool create` result immediately returns an error and prevents `zfs create` from running. The existing `.context("Failed to create zpool")?` then stops the setup path, while the command's stderr remains visible in the service logs.

Therefore, this change does not fix the stated error-propagation bug; it only changes how the diagnostic is formatted and renames the dataset-creation context. That does not justify the additional manual `Command::output()` implementation.

PR #866, opened 2026-07-31
fix(gateway): align health routes

kvinwang:
## Problem

Health and compatibility routes were mounted inconsistently across Gateway listeners. Probes using the documented paths received 404 responses even while the service was healthy.

## Root cause and fix

Mount health handlers under the intended stable paths and keep route separation explicit on the main, admin, and debug listeners.

The public RPC route intentionally keeps `trim: "Tproxy."`: normal Gateway RPC methods are unprefixed (for example, `/prpc/Info`), while the `Tproxy.` prefix is retained only for legacy-client compatibility. This PR does not add or require a `Gateway.` URL prefix.

## Implementation

- Mount `/health` on the main, admin, and debug listeners.
- Add the `/health/dashboard` compatibility alias on all three listeners.
- Fall back to a minimal readiness page if the full dashboard cannot render.

Changed paths:

- `dstack/gateway/src/main.rs`
- `dstack/gateway/src/web_routes.rs`

## Scope

This PR addresses one logical `gateway` routing finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

## Verification

- `git diff --check origin/master..origin/codex/fix-gateway-rpc-health-routes`: passed.

### DIFF
--- a/dstack/gateway/src/main.rs
+++ b/dstack/gateway/src/main.rs
@@ -325,6 +325,7 @@ async fn main() -> Result<()> {
     let mut rocket = rocket::custom(figment)
         .mount("/prpc", prpc!(Proxy, RpcHandler, trim: "Tproxy."))
         .mount("/", web_routes::health_routes())
+        .mount("/", web_routes::dashboard_alias_routes())
         // Mount WaveKV sync endpoint (requires mTLS gateway auth)
         .mount("/", web_routes::wavekv_sync_routes())
         .attach(AdHoc::on_response("Add app version header", |_req, res| {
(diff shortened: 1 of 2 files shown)


kvinwang:
Closing because the `/health/dashboard` route was invented by the acceptance fixture rather than being part of the Gateway API contract. PR #841 now uses the existing main-listener `/health` endpoint for liveness and the authenticated Admin `/` route for the dashboard. The legacy `Tproxy.` RPC compatibility behavior remains unchanged.

PR #887, opened 2026-07-31
fix(guest): disable active swap before replacement

kvinwang:
## Problem

The guest attempted to replace or format a swap device while it could still be active. The operation then failed with a busy device and could modify storage still used by the kernel.

## Root cause and fix

Disable swap and verify the device is inactive before replacement, then recreate and reactivate it through the normal flow.

## Implementation

- `fix(guest): disable active swap before replacement`
- `fix(guest): compare resolved swap paths safely`

Changed paths:

- `dstack/dstack-util/src/system_setup.rs`

## Scope

This PR addresses one logical `guest` finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

## Verification

- `git diff --check origin/master..origin/codex/fix-guest-swap-replacement`: passed.

### DIFF
--- a/dstack/dstack-util/src/system_setup.rs
+++ b/dstack/dstack-util/src/system_setup.rs
@@ -2211,6 +2211,31 @@ impl<'a> Stage0<'a> {
         }
     }
 
+    fn active_swap_path(swap_path: &Path) -> Result<Option<PathBuf>> {
+        let canonical_swap_path = fs::canonicalize(swap_path)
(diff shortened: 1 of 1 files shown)


kvinwang:
Closing because the reported condition requires rerunning guest setup while its swap is already active; normal boot starts without active swap inherited from the previous boot. The related acceptance case in #841 was also corrected to validate the real boot/reboot lifecycle instead of asserting the proposed source implementation.

PR #902, opened 2026-07-31
[STACKED on #883] fix(util): stage certificate and key together

kvinwang:
> **STACKED PR — merge #883 first.**

## Problem

Certificate and private key were replaced as separate writes. A crash or reader between the writes could observe a new certificate paired with the old key, causing TLS loading and handshakes to fail.

## Root cause and fix

Stage the matching certificate/key outputs together and publish the pair through an atomic commit sequence.

## Implementation

- `fix(util): stage certificate and key together`

Changed paths:

- `dstack/crates/dstack-cli-core/src/fsutil.rs`
- `dstack/dstack-util/src/main.rs`

## Scope

This PR addresses one logical `util` finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

- Parent PR: #883
- GitHub base branch: `codex/fix-guest-atomic-credentials`
- Merge the parent first; this PR contains only the additional delta above that parent.

## Verification

- `git diff --check origin/codex/fix-guest-atomic-credentials..origin/codex/fix-util-atomic-cert-key`: passed.

### DIFF
--- a/dstack/Cargo.lock
+++ b/dstack/Cargo.lock
@@ -2208,6 +2208,7 @@ dependencies = [
  "curve25519-dalek",
  "dcap-qvl",
  "dstack-attest",
+ "dstack-cli-core",
  "dstack-gateway-rpc",
  "dstack-kms-rpc",
  "dstack-types",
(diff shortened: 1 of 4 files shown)


kvinwang:
Closing because staging two files and renaming them sequentially does not make the certificate/key pair atomic: a crash, reader, or second rename failure can still observe a new certificate with the old key. The related #841 case was updated to validate certificate/key binding, permissions, individual output failures, and retry without asserting unsupported two-file transaction semantics.

PR #893, opened 2026-07-31
fix(supervisor): reject untrusted client sockets

kvinwang:
## Problem

Supervisor accepted commands from any process able to connect to its Unix socket. If filesystem permissions were misconfigured, an untrusted local process could issue privileged lifecycle operations.

## Root cause and fix

Check peer credentials on accepted Unix connections and reject clients outside the trusted identity boundary.

## Implementation

The branch records the following focused implementation work:

- `fix(supervisor): reject untrusted client sockets`
- `build(supervisor): lock client libc dependency`
- `fix(supervisor): retain socket path for validation`
- `test(supervisor): add socket fixture dependency`
- `test(supervisor): expose trusted socket rejection`
- `test(supervisor): secure trusted socket directory`

Changed paths:

- `dstack/Cargo.lock`
- `dstack/supervisor/client/Cargo.toml`
- `dstack/supervisor/client/src/lib.rs`

## Scope

This PR addresses one logical `supervisor` finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

This PR is based directly on `master` and does not require another split product PR to merge first.

## Verification

- `git diff --check origin/master..origin/codex/fix-supervisor-trusted-client-sockets`: passed.
- The declared base was verified as an ancestor of the PR head.
- Focused compile/check verification was run for the changed component where applicable; non-Rust packaging or configuration changes were reviewed against their exact branch delta.

PR #905, opened 2026-07-31
feat(supervisor): expose concurrent UDS auto-start

kvinwang:
## Summary

Expose the Supervisor client's existing Unix-socket auto-start flow through the CLI and make concurrent starters converge on one daemon.

## Changes

- Add CLI options for auto-starting a missing Supervisor over a Unix socket.
- Serialize concurrent starts with an owner-only lock and re-probe after acquiring it.
- Start the Supervisor with `umask 077`, so its runtime files and socket are not group/world writable.
- Keep structured command output on stdout by sending tracing output to stderr.
- Return the shutdown response before notifying Rocket to stop.

This branch is rebased directly onto `master`. It does not include the client-side socket path validation from the closed PR #893; normal Unix socket and parent-directory permissions remain the access-control boundary.

## Verification

- `cargo check --manifest-path dstack/Cargo.toml -p supervisor-client -p supervisor`
- `cargo test --manifest-path dstack/Cargo.toml -p supervisor-client -p supervisor`
- `cargo fmt --manifest-path dstack/Cargo.toml --all`
- `git diff --check`

### DIFF
--- a/dstack/Cargo.lock
+++ b/dstack/Cargo.lock
@@ -7405,6 +7405,7 @@ dependencies = [
  "hyper",
  "hyper-util",
  "hyperlocal",
+ "libc",
  "log",
  "serde",
  "serde_json",
--- a/dstack/supervisor/client/Cargo.toml
+++ b/dstack/supervisor/client/Cargo.toml
@@ -27,6 +27,7 @@ serde.workspace = true
 http-body-util.workspace = true
 tracing-subscriber.workspace = true
 log.workspace = true
+libc.workspace = true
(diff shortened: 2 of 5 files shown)


kvinwang:
Closing the concurrent UDS auto-start feature because the supported deployment has one VMM owner per runtime directory. The two independent fixes were split into #994 (machine-readable client output) and #995 (graceful shutdown response). PR #841 was updated to test the client lifecycle directly without requiring CLI auto-start or concurrent-starter behavior.

PR #924, opened 2026-07-31
fix(certbot): pace and stop the daemon cleanly

kvinwang:
## Problem

Certbot loop retries could run without proper pacing and shutdown did not interrupt the wait/work cleanly. This caused provider request bursts and delayed process termination.

## Root cause and fix

Apply the configured interval through an interruptible wait and honor shutdown before starting another reconciliation.

## Implementation

- `fix(certbot): pace daemon and run once hook`
- `fix(certbot): stop daemon gracefully`

Changed paths:

- `dstack/certbot/cli/src/main.rs`
- `dstack/certbot/src/bot.rs`

## Scope

This PR addresses one logical `certbot` finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

## Verification

- `git diff --check origin/master..origin/codex/fix-certbot-daemon-lifecycle`: passed.

### DIFF
--- a/dstack/certbot/cli/src/main.rs
+++ b/dstack/certbot/cli/src/main.rs
@@ -154,8 +154,26 @@ async fn renew(config: &PathBuf, once: bool, force: bool) -> Result<()> {
     if once {
         bot.renew_and_run_hook(force).await?;
     } else {
-        bot.run().await;
+        tokio::select! {
+            _ = bot.run() => unreachable!("certbot daemon returned"),
+            result = shutdown_signal() => result?,
+        }
     }
+    Ok(())
+}
+
+async fn shutdown_signal() -> Result<()> {
+    #[cfg(unix)]
+    {
+        let mut terminate =
+            tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate())?;
+        tokio::select! {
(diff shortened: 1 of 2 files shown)


kvinwang:
Post-merge correction to the root-cause description, so future archaeology on this commit is not misled.

The PR describes the symptom as "provider request bursts". The idle path never reaches the provider. `renew_cert_if_needed` returns early on a local check:

```rust
// dstack/certbot/src/acme_client.rs
if !need_renew(cert_pem, expires_in)? {
    return Ok(None);
}
```

`need_renew` only parses the on-disk certificate, and `create_cert_if_needed` above it is a local file check too. So in the steady state the old `continue` — which skipped `sleep(renew_interval)` on exactly the common `!renewed` path — produced a hot spin loop: a file read, an x509 parse and two `info!` lines per iteration, at full CPU, indefinitely. No ACME or Cloudflare traffic was involved; the error path was the one that stayed paced, since it fell through to the sleep.

The fix is correct as merged, and the CPU/log-flood impact is arguably worse than the described one for a CVM. Only the stated cause was off.

Follow-ups filed: #1012, #1013, #1014.

PR #955, opened 2026-07-31
[STACKED on #952] feat(kms): return configured historical root keys

kvinwang:
> **STACKED PR — merge #952 first.**

## Problem

KMS retained historical root keys for compatibility/recovery but its API returned only the active root. Guests needing to decrypt or verify state created under an older configured root could not obtain it.

## Root cause and fix

Return the configured historical roots together with the active key under the explicit historical-key response contract.

## Implementation

- `feat(kms): return configured historical root keys`
- `test(kms): cover historical root key inventory`
- `docs(kms): define cold backup recovery procedure`

Changed paths:

- `dstack/kms/README.md`
- `dstack/kms/kms.toml`
- `dstack/kms/src/config.rs`
- `dstack/kms/src/main_service.rs`

## Scope

This PR addresses one logical `kms` finding. It intentionally excludes the acceptance-test infrastructure from #841 and unrelated product fixes from #840.

## Dependency and merge order

- Parent PR: #952
- GitHub base branch: `codex/fix-kms-ca-restart`
- Merge the parent first; this PR contains only the additional delta above that parent.

## Verification

- `git diff --check origin/codex/fix-kms-ca-restart..origin/codex/feat-kms-historical-root-keys`: passed.

### DIFF
--- a/dstack/kms/README.md
+++ b/dstack/kms/README.md
@@ -232,3 +232,21 @@ The `SignCert` RPC is used by the dstack app to sign a TLS certificate. In this 
 - Verify the CSR signature
 - Query the smart contract to check if the app is authorized
 - If authorized, sign the CSR with the CA root key and return the certificate chain to the app
+
+## Cold backup and recovery
+
+KMS root material is backed up as a complete, offline copy of `core.cert_dir`.
+There is no online backup RPC. Stop the KMS before taking or restoring a copy,
+preserve file modes and ownership, and protect the backup with the same controls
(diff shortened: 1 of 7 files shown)


kvinwang:
Closing because KMS does not currently implement a complete root-key rotation flow. This change only returns configured historical keys, while the onboarding client still requires exactly one key and neither persists nor uses historical keys, so configuring them would break onboarding rather than provide an end-to-end recovery path.

PR #1004, opened 2026-08-05
feat(gateway): safely rotate ACME credentials

kvinwang:
## Problem

A corrupt persisted ACME credential must not take down Gateway or discard certificates that are already serving traffic. Operators also need a controlled way to replace the shared ACME account and its account-bound CAA records.

## Solution

- Reject malformed ACME credential JSON for the current issuance/renewal attempt without affecting Gateway startup, TLS serving, or other certificate state.
- Add the admin-only `RotateAcmeCredentials` RPC.
- Create the replacement ACME account before changing shared state.
- Update `issue` and `issuewild` CAA records for every configured ZT domain, using each domain's configured DNS credential.
- Publish the replacement credential to WaveKV only after all CAA updates succeed.
- Never return private ACME credential material through the API.
- Serialize rotations handled by one node. WaveKV 1.0 has no CAS, so operators must not invoke rotation concurrently through multiple gateway nodes.

Gateway does not retain a long-lived ACME client. Each issuance or renewal reloads credentials from WaveKV, so nodes automatically use the replacement credential on their next attempt after synchronization.

## Failure isolation

CAA/account creation failures abort rotation without replacing the persisted credential. Existing certificates remain loaded and continue serving traffic. Corrupt credentials fail only the affected ACME operation and can be repaired through the rotation API.

## Verification

- `cargo check -p dstack-gateway`
- `cargo test -p dstack-gateway credential_tests -- --nocapture`
- `git diff --check`

### DIFF
--- a/dstack/gateway/rpc/proto/gateway_rpc.proto
+++ b/dstack/gateway/rpc/proto/gateway_rpc.proto
@@ -164,6 +164,14 @@ message AcmeInfoResponse {
   string account_attestation = 5;
 }
 
+// Result of replacing the shared ACME account credentials.
+message RotateAcmeCredentialsResponse {
+  // URI of the newly-created ACME account. The private credentials are never returned.
+  string account_uri = 1;
+  // Number of ZT domains whose CAA records were updated for the new account.
+  uint32 domains_updated = 2;
+}
+
 // Get HostInfo for associated instance id.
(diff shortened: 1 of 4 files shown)


kvinwang:
Superseded by the restored and unstacked #935.

PR #1042, opened 2026-08-11
fix(os): converge both guest images on one nftables netfilter path

### LINKED ISSUE #1032: Incus managed networking compatibility gaps observed on DStack 0.5.9
## Summary

We run Incus system containers inside a DStack CVM. On the released DStack `0.5.9` image (`6.9.0-dstack`), standard Incus managed-bridge paths encounter several missing netfilter capabilities.

This is a released-image compatibility report, not a claim that ordinary DStack Compose applications are broken. We reviewed current `next` at `312dc605ebaaf93372534ffb135ea011533d28e9`, but have not boot-tested that unreleased source and do not claim it reproduces these runtime failures.

The latest stable guest release is `v0.5.11`. Compared with `v0.5.9`, its relevant kernel fragment only adds DirtyFrag hardening, while `v0.5.10` added the `nft` userspace package. That may change Incus firewall-driver selection, but the release does not add the named kernel symbols or matching ebtables module packages. We have not run the same probes on `v0.5.11`, so this issue asks for its status rather than asserting it still fails.

## Environment

- DStack: `0.5.9`
- Kernel: `6.9.0-dstack`
- Incus: `6.0.6`
- Docker owns an active legacy IPv4 ruleset with `FORWARD DROP`

### 1. The default DHCP bridge path requires a missing CHECKSUM target

Incus attempted:

```text
iptables -w -t mangle -I POSTROUTING -o incusbr0 -p udp --dport 68 -j CHECKSUM --checksum-fill
```

It failed with:

```text
Warning: Extension CHECKSUM revision 0 not supported, missing kernel module?
iptables: No chain/target/match by that name.
```

The captured kernel configuration had no `CONFIG_NETFILTER_XT_TARGET_CHECKSUM`. Incus' Xtables [`NetworkSetup`](https://github.com/lxc/incus/blob/5231e7a1beca905d57208b441040f9ebdc6a2c6f/internal/server/firewall/drivers/drivers_xtables.go#L438-L471) adds this target when IPv4 DHCP access is enabled.

### 2. The Xtables cleanup path cannot list the legacy IPv6 NAT table

Bridge creation still failed with DHCP disabled, no IPv6 address, and no address forwards:

```text
Failed applying firewall address forwards: Failed to list IPv6 rules (table nat)
```

Incus calls [`forwardSetupFirewall`](https://github.com/lxc/incus/blob/5231e7a1beca905d57208b441040f9ebdc6a2c6f/internal/server/network/driver_bridge.go#L2019-L2023) even for an empty forward set, and Xtables [`clearNetworkForwards`](https://github.com/lxc/incus/blob/5231e7a1beca905d57208b441040f9ebdc6a2c6f/internal/server/firewall/drivers/drivers_xtables.go#L1527-L1567) loops over IPv4 and IPv6.

That unconditional family-6 cleanup is Incus behavior. The DStack image gap it exposes is narrower: legacy `ip6tables` cannot list the NAT table. The same host successfully services that query through Debian's real `ip6tables-nft` shim.

### 3. The legacy ebtables bridge-filter path has no usable kernel support

Enabling the Incus bridge-filter path failed with:

```text
modprobe: FATAL: Module ebtables not found in directory /lib/modules/6.9.0-dstack
The kernel doesn't support the ebtables 'filter' table.
```

The captured configuration had `CONFIG_BRIDGE_NF_EBTABLES=m`, but the running image had no usable matching module tree, and the ebtables filter-table options were not enabled. Incus' Xtables [`InstanceSetupBridgeFilter`](https://github.com/lxc/incus/blob/5231e7a1beca905d57208b441040f9ebdc6a2c6f/internal/server/firewall/drivers/drivers_xtables.go#L808-L849) executes ebtables rules on this path; installing userspace alone cannot provide the missing kernel tables.

## Secondary bridge-filter observation

An earlier two-container probe with Incus `6.0.0` also found the nftables `bridge` family unavailable. Incus renders per-instance filters in that family in [`Nftables.InstanceSetupBridgeFilter`](https://github.com/lxc/incus/blob/v6.0.0/internal/server/firewall/drivers/drivers_nftables.go#L396-L457). We retained the canary conclusion but not its exact stderr, so this is included as supporting context rather than presented as a fourth equally evidenced failure. We can rerun it on a disposable CVM.

## Current workaround and impact

Our current configuration:

- disables Incus DHCP and IPv6 and injects static addresses;
- uses legacy iptables for IPv4 and `ip6tables-nft` for the unavoidable IPv6 cleanup;
- does not request Incus MAC/IP filtering or ACLs; and
- uses the native bridge-port isolation flag for basic east-west isolation.

This avoids wrappers, ignored errors, custom firewall rules, locally built modules, and an Incus fork. It also means the normal Incus filtering features are unavailable, and we make no anti-spoof or destination/port ACL claim.

## Request

Could DStack provide or identify one released netfilter path that supports Incus managed bridges, including:

- the CHECKSUM target used by the DHCP path;
- coherent IPv4/IPv6 NAT behavior alongside DStack's Docker firewall; and
- one complete bridge-filtering path, preferably nftables `bridge`, or the full matching ebtables kernel support and module packages?

If a newer image already provides these capabilities, please identify the first image version and whether each capability is built in or shipped as a loadable module. We can rerun the exact bridge and two-container probes on a disposable CVM. We can also provide the retained configuration values, command output, and failed service state from the zero-capacity test node; a rerun can capture and attach the complete generated configuration.

## Responsibility boundary

Incus' empty-set IPv6 cleanup and its failure-path dnsmasq/INPUT-rule residue are relevant context but are not being reported as DStack defects. This issue is limited to capabilities missing from the released DStack guest image.

kvinwang:
Closes #1032.

## Problem

Incus managed bridges fail inside a dstack CVM in three places. The reporter hit these on `0.5.9`; all three are still reproducible on `next`.

Two causes sit underneath them.

**A kconfig fragment line is a request, not a guarantee.** Kconfig silently drops a symbol whose dependencies are unmet, and silently clamps a tristate to the value of what it depends on. Nothing on the Yocto side checked, so `dstack-docker.cfg` accumulated five lines no build ever produced — including `CONFIG_BRIDGE_NF_EBTABLES=y` under `CONFIG_BRIDGE=m`, which is directly why `modprobe ebtables` finds nothing.

**The two guest images ran different netfilter dataplanes, and nothing said so.** OE's iptables recipe builds only the legacy binaries unless `libnftnl` is in `PACKAGECONFIG`, so on Yocto every rule — Docker's, dstack's own `DSTACK_WG` chain, a tenant's — landed in the legacy tables. mkosi meanwhile programs nftables, because Debian's iptables defaults to the nft frontend. Neither image pinned a backend; each inherited its distribution's default, and the defaults differ.

That is the mechanism behind the bug report. Incus selects its nftables firewall driver if any nftables rule exists, and otherwise falls back to xtables *to avoid mixing drivers*. So Docker's legacy rules alone steered it onto the xtables path — and all three reported failures live in that path's managed-bridge code. The legacy and nf_tables rulesets register at the same hooks but cannot see each other, so the effective ruleset was only knowable by querying both.

Measured on real artifacts, the capability gaps were disjoint:

| Capability | Yocto before | mkosi before |
| --- | --- | --- |
| `NETFILTER_XT_TARGET_CHECKSUM` | present | **missing** |
| `NF_TABLES_BRIDGE` | present | **missing** |
| `IP6_NF_IPTABLES` | **missing** | present |
| ebtables tables | **missing** | **missing** |

## Fix

Converge both images on nftables, and give both the bridge-filtering capability that path needs.

Ten commits, one concern each:

1. **Correct the five unreachable assertions** in `dstack-docker.cfg`. No built kernel changes — each line now records what the build already produced.
2. **Yocto kernel capabilities** — the nftables bridge family, the ebtables matches `nft_compat` loads, `CHECKSUM`, and the IPv6 tables.
3. **Yocto rootfs packaging** — the image installs only explicitly named `kernel-module-*` packages, so the bridge modules are added.
4. **mkosi kernel capabilities** — the same bridge-filtering block, symbol for symbol.
5. **Share the kernel-config gate** between both backends.
6. **`iptables -w`** in the guest WireGuard setup.
7. **Drop the WireGuard config when its rules were not applied.**
8. **Program netfilter through nftables on Yocto** — `libnftnl`, plus explicit links for the frontends the recipe leaves behind.
9. **Assert the frontend on mkosi** rather than inheriting Debian's default.
10. **Docs.**

### Why the ebtables answer changed shape

The obvious fix for failure 3 is to enable the legacy ebtables tables. This PR deliberately does not: once both images are on nftables, Incus takes its nftables driver, which renders bridge filters in the nft `bridge` family and never invokes the legacy tables. So the fix is `NF_TABLES_BRIDGE` plus the `ebt_*` match modules that `nft_compat` still loads — and `BRIDGE_NF_EBTABLES_LEGACY` stays off on both backends.

That also settles where the `ebtables` command comes from. It is `xtables-nft-multi` dispatching on `argv[0]`, not the standalone `ebtables` recipe — that one is legacy-only and registers `/usr/sbin/ebtables` through `update-alternatives`, so installing both would collide on the same path.

The two fragments' bridge-filtering blocks are now identical, which is the point: that is the capability the images silently disagreed on.

### The two `dstack-util` commits

Pre-existing bugs found while tracing the only dstack code that programs rules inside the CVM. Unrelated to Incus.

`DSTACK_WG` restricts the WireGuard listen port to the gateway peers. It shells out to `iptables` with no `-w`, so it does not wait for `/run/xtables.lock` — it exits immediately with *"another app is currently holding the xtables lock"*. dockerd rewrites its own rules on every container start and stop, which is exactly what happens during boot when this runs. Losing that race leaves the chain flushed or half-populated while its jump from `INPUT` is already installed, so packets fall through and the chain stops restricting anything. WireGuard is still cryptographically authenticated, so this is defence in depth rather than the trust boundary — but it is the chain's only purpose.

It also does not self-heal. The rendered config is written *before* the rules are programmed, and is also what the refresher compares against to decide whether to act, so a failure leaves the file claiming a configuration that was never applied and every later refresh takes the `WireGuard config unchanged, skipping reconfiguration` early return. Commit 7 makes the file mean "this was applied". Removing it on failure is safe because the next attempt rewrites it before the `wg-quick down` that reads it, and correct because `/etc` is a tmpfs overlay — the file already shares the reboot lifetime of the rules it stands for.

## Verification

**No guest image was built or booted, and the bitbake recipe changes are untested.** What was verified:

- **Kernel config, reproduced exactly.** Ran the mkosi build's own steps — `x86_64_defconfig`, `merge_config.sh -m`, `olddefconfig` — against Linux 6.18.40 source, the pinned `KERNEL_VERSION` with the matching `KERNEL_SHA256`.
- **Yocto side against a built artifact.** Used a deployed `kernel-config` from a real build (linux-yocto 6.18.24) as the baseline. Every assertion satisfied after; fourteen missing before.
- **Both backends' bridge blocks now diff clean** against each other.
- **Kconfig dependencies read from source, not assumed.** This mattered repeatedly: `NETFILTER_XT_TARGET_CHECKSUM` depends on `IP_NF_MANGLE || IP6_NF_MANGLE || NFT_COMPAT`, and `NFT_COMPAT` was already set, so it needed one line rather than a legacy-mangle chain. `BRIDGE_EBT_TABLE_FILTER`/`_NAT` **do not exist** — the symbols are `BRIDGE_EBT_T_FILTER`/`_T_NAT`. `BRIDGE_NF_EBTABLES_LEGACY` is a tristate capped at `=m` by `BRIDGE=m`, not the `=y` its name suggests. `NF_TABLES_BRIDGE` builds no module of its own, so there is no `kernel-module-nf-tables-bridge` to package.
- **The frontend links.** Confirmed `xtables-nft-multi` accepts every linked name as a subcommand, and that invoking it as `ebtables` reports `ebtables 1.8.10 (nf_tables)`.
- **The gate, four cases.** Passes on both final configs; fails on the pre-patch Yocto kernel-config naming exactly the capabilities this series adds; fails on a truncated fragment; fails with no fragment.
- **The parity assertion, three cases.** Accepts an all-nft rootfs; reports `resolves to xtables-legacy-multi, wanted xtables-nft-multi` when the alternative is repointed; and rejects a dangling link whose binary was removed. Symlinks are followed *inside* the image rather than with `os.path.realpath()`, which would escape into the build host — a legacy-only rootfs passed until that was fixed, because the host happened to have an `xtables-nft-multi` at the absolute path the image's symlink named.
- **REUSE lint** passes; moving the gate script out of `os/mkosi/**` had taken it out of the blanket annotation that covered it.
- **Rust.** `cargo build`, `clippy --all-features`, `fmt --check` clean. Confirmed `cmd_lib` splits `-w $xtables_wait` into two argv entries by running the same macro against `printf`. shellcheck and `json.tool` clean.

The reporter offered to rerun their bridge and two-container probes on a disposable CVM. That is the step this PR cannot substitute for.

## Follow-ups, deliberately not here

- **`dstack.cfg` is not gated yet.** Six lines the build does not satisfy: `HOTPLUG_CPU`, `SCSI` and `INPUT` are forced back on by machine-level kernel features, and `TLS`, `CRYPTO_GCM` and `CRYPTO_CHACHA20POLY1305` do not come out as asserted despite a detailed comment explaining why kTLS needs them. Each needs its own decision.
- **SELinux is off on Yocto, on for mkosi.** `CONFIG_SECURITY_SELINUX=y` was asserted but never took effect — the linux-yocto tiny tree leaves `CONFIG_SECURITY` off. Enabling it changes the kernel's measurements, so commit 1 documents it rather than silently turning it on.
- **`DSTACK_WG` is IPv4-only.** No `ip6tables` counterpart, so a gateway endpoint reachable over IPv6 is not filtered.
- **`ephemeral-docker.sh` starts a dockerd without `--iptables=false`**, so it programs rules in the guest while it runs.


kvinwang:
Thanks — three of the five are addressed in 82f11ce93; one is a false positive.

**Addressed**

- `check-kernel-config.sh` unquoted `$fragment` — quoted. Note `[[ ]]` does not word-split or glob-expand its operands, so this was not reachable as a bug, but the inconsistency with the quoted uses two lines down is worth removing on its own. Confirmed a fragment path containing a space is still handled.
- The `-w` budget comment — reworded. It now states the aggregate explicitly: the bound is per invocation, so with four fixed calls plus one per peer a lock held throughout costs roughly `5s * (4 + peers)`, which is the thing that has to stay inside the 600s `WatchdogSec`.
- `apply_wg_config`'s doc comment — reworded. You are right that kernel state can be partly modified before the error returns; the guarantee is about the *retry*, not about the state being untouched. It now says so: best-effort, and what the caller ensures is that the next refresh redoes the sequence because the config file this wrote is removed.

**Not applying: the `kernel-module-nft-*-bridge` package names**

Both comments propose deriving these from the Kconfig symbol (`CONFIG_NFT_BRIDGE_META` → `kernel-module-nft-bridge-meta`). OpenEmbedded derives them from the **module filename**, not the symbol, and the two differ here — the symbol reads `BRIDGE_META` while the object is `meta_bridge`:

```
# linux 6.18.40, net/bridge/netfilter/Makefile
obj-$(CONFIG_NFT_BRIDGE_META)   += nft_meta_bridge.o
obj-$(CONFIG_NFT_BRIDGE_REJECT) += nft_reject_bridge.o
```

The naming path is `kernel-module-split.bbclass:226-240`:

```
module_regex   = r'^(.*)\.k?o(?:\.(gz|xz|zst))?$'    # captures the filename stem
module_pattern = <kernel_package_name> + '-module-%s'
do_split_packages(..., file_regex=module_regex, output_pattern=module_pattern, ...)
```

and `do_split_packages` runs each captured name through `oe.package.legitimize_package_name`, which is `s.lower().replace('_', '-')` (`meta/lib/oe/package.py:628-642`).

So `nft_meta_bridge.ko` → `kernel-module-nft-meta-bridge` and `nft_reject_bridge.ko` → `kernel-module-nft-reject-bridge`, which is what the recipe now names. The comparison to `nft_reject_inet` → `kernel-module-nft-reject-inet` actually supports this: that one matches its *filename* too; it is only indistinguishable from the symbol-derived name because for that module the two happen to coincide.

A Yocto build is running now and will confirm this empirically — a wrong package name fails the rootfs assembly outright, so it is not a silent failure mode either way.

kvinwang:
## mkosi backend built and tested

`make os-image-mkosi` (containerised repro build) completed clean, kernel compiled cold rather than served from cache. `os_image_hash a1ab5df2ef7eaefd3a8bc5f8cc61eee2a75247fb1dec5d372dca31160e4a6f1f`.

### The three reported failures, reproduced as successes

Run against the **shipped rootfs** (unsquashed from `rootfs.img.parted.verity`), chrooted into a private network namespace, using the issue's own commands:

**Failure 1 — the CHECKSUM target**
```
# iptables -w 5 -t mangle -I POSTROUTING -o incusbr0 -p udp --dport 68 -j CHECKSUM --checksum-fill
exit=0

Chain POSTROUTING (policy ACCEPT 0 packets, 0 bytes)
 pkts bytes target     prot opt in     out       source       destination
    0     0 CHECKSUM   udp  --  *      incusbr0  0.0.0.0/0    0.0.0.0/0    udp dpt:68 CHECKSUM fill
```

**Failure 2 — listing the IPv6 nat table**
```
# ip6tables -t nat -L
Chain PREROUTING (policy ACCEPT)
Chain INPUT (policy ACCEPT)
...
```

**Failure 3 — the ebtables filter table**
```
# ebtables --version
ebtables v1.8.11 (nf_tables)
# ebtables -L
Bridge table: filter

Bridge chain: INPUT, entries: 0, policy: ACCEPT
Bridge chain: FORWARD, entries: 0, policy: ACCEPT
```

### One ruleset, not two

The same mangle rule seen through `nft`, which is the point of the convergence commit — the `iptables` invocation above landed in nftables, not in a legacy table only `iptables` can see:

```
# nft list table ip mangle
	chain POSTROUTING {
		type filter hook postrouting priority mangle; policy accept;
		oifname "incusbr0" udp dport 68 counter packets 0 bytes 0 xt target "CHECKSUM"
	}
```

Frontend resolution in the shipped rootfs, following the full chain:

```
iptables  -> /etc/alternatives/iptables  -> iptables-nft  -> xtables-nft-multi
ip6tables -> /etc/alternatives/ip6tables -> ip6tables-nft -> xtables-nft-multi
ebtables  -> /etc/alternatives/ebtables  -> ebtables-nft  -> xtables-nft-multi
```

### Shipped kernel

Config extracted from the released `bzImage` via `extract-ikconfig` (Linux/x86 6.18.40), then run through the repo's own gate — **pass**:

| | |
| --- | --- |
| `NF_TABLES_BRIDGE`, `NFT_BRIDGE_META`, `NFT_BRIDGE_REJECT` | `m` |
| `NETFILTER_XT_TARGET_CHECKSUM` | `m` |
| `BRIDGE_EBT_ARP/IP/IP6/AMONG/LIMIT/VLAN` | `m` |
| `BRIDGE_NF_EBTABLES_LEGACY`, `BRIDGE_EBT_T_FILTER`, `NETFILTER_XTABLES_LEGACY` | absent, as intended |

Modules present in the rootfs: `nft_meta_bridge.ko`, `nft_reject_bridge.ko`, `xt_CHECKSUM.ko`, `ebt_{arp,ip,ip6,among,limit,vlan}.ko`. Legacy `ebtables.ko` / `ebtable_filter.ko` / `ebtable_nat.ko` absent, as intended.

The new parity assertion ran as part of the build and passed (`Yocto functional parity paths accepted`) — so the frontend resolution above is now enforced, not just observed.

### Scope of this test

The chroot exercised the shipped **userspace** against the **build host's** kernel, not the shipped 6.18.40-dstack one; the shipped kernel's capabilities were verified separately from its own embedded config, above. A full boot test of the CVM image is not covered here.

The Yocto build is still running on a separate machine; results to follow, including whether the `kernel-module-nft-meta-bridge` package name resolves.

kvinwang:
## Yocto backend built and tested

`make os-image` on a 32-core host, `Tasks Summary: Attempted 6883 tasks ... all succeeded`, `rc=0`, **0 `ERROR:` lines**. Release tarballs produced; `export-artifacts.sh` — which this PR extends with the kernel-config gate — ran and emitted the manifest.

Both guest images are now built and tested. All three reported failures reproduce as successes on **both**.

### The package-name question is settled

`recipe dstack-rootfs-1.0-r0: task do_rootfs: Succeeded`, no `Nothing RPROVIDES` anywhere in the log, and the modules are in the image:

```
/lib/modules/6.18.24-dstack/kernel/net/bridge/netfilter/nft_meta_bridge.ko
/lib/modules/6.18.24-dstack/kernel/net/bridge/netfilter/nft_reject_bridge.ko
/lib/modules/6.18.24-dstack/kernel/net/netfilter/xt_CHECKSUM.ko
```

So `kernel-module-nft-meta-bridge` resolves, as derived from the module filename. Had the names been taken from the Kconfig symbol (`kernel-module-nft-bridge-meta`), `do_rootfs` would have failed outright.

### The bbappend works

Frontends in the shipped Yocto rootfs — the recipe's own `do_install` creates the first two once `libnftnl` is enabled, the rest are the ones this PR's `iptables_%.bbappend` adds:

```
iptables          -> /usr/sbin/xtables-nft-multi
iptables-save     -> /usr/sbin/xtables-nft-multi
ip6tables         -> /usr/sbin/xtables-nft-multi
ip6tables-restore -> /usr/sbin/xtables-nft-multi
ebtables          -> /usr/sbin/xtables-nft-multi
ebtables-save     -> /usr/sbin/xtables-nft-multi
```

No standalone legacy `ebtables` binary, so the `update-alternatives` collision the commit message predicts does not arise. `iptables-legacy` / `ip6tables-legacy` remain available as the suffixed names, which is what the docs describe.

### The three failures, on the shipped Yocto rootfs

Chrooted into a private network namespace:

```
# ebtables --version
ebtables v1.8.13 (nf_tables)

# ebtables -L                          <- failure 3
Bridge table: filter
Bridge chain: INPUT, entries: 0, policy: ACCEPT

# ip6tables -t nat -L                  <- failure 2
Chain PREROUTING (policy ACCEPT)

# iptables -w 5 -t mangle -I POSTROUTING -o incusbr0 -p udp --dport 68 -j CHECKSUM --checksum-fill
exit=0                                 <- failure 1
CHECKSUM   udp  --  0.0.0.0/0   0.0.0.0/0   udp dpt:68 CHECKSUM fill
```

### Kernel

Freshly built `kernel-config` (Linux/x86 6.18.24), checked with the gate this PR adds — **pass**. `NF_TABLES_BRIDGE`, `NFT_BRIDGE_META`, `NETFILTER_XT_TARGET_CHECKSUM`, `IP6_NF_IPTABLES/NAT/FILTER/MANGLE` and the `BRIDGE_EBT_*` matches all `=m`; `BRIDGE_NF_EBTABLES_LEGACY` and `BRIDGE_EBT_T_FILTER` absent, as intended. `ip6_tables.ko`, `ip6table_nat.ko` and `ip6table_filter.ko` are in the image via the `iptables` recipe's existing `RRECOMMENDS`, which resolve now that the kernel builds them.

### Both backends agree

| | Yocto 6.18.24 | mkosi 6.18.40 |
| --- | --- | --- |
| `ebtables` reports | `v1.8.13 (nf_tables)` | `v1.8.11 (nf_tables)` |
| Frontends resolve to | `xtables-nft-multi` | `xtables-nft-multi` |
| `NF_TABLES_BRIDGE` / `NFT_BRIDGE_META` / `NFT_BRIDGE_REJECT` | `m` | `m` |
| `NETFILTER_XT_TARGET_CHECKSUM` | `m` | `m` |
| `BRIDGE_EBT_*` matches | `m` | `m` |
| `BRIDGE_NF_EBTABLES_LEGACY`, `BRIDGE_EBT_T_*` | absent | absent |
| Kernel-config gate | pass | pass |

The offline Kconfig reproduction quoted in the PR description turned out to match both real builds symbol for symbol.

### Still not covered

Neither image was booted. The chroot tests exercised the shipped **userspace** against each build host's kernel; each shipped kernel's capabilities were verified from its own config (extracted from `bzImage` for mkosi, from the deployed `kernel-config` for Yocto). An end-to-end Incus test inside a real CVM remains the reporter's offer in #1032, and is the one thing that would close this fully.

kvinwang:
## Tested in a real TDX CVM

The earlier results were artifact inspection plus a chroot against the *build host's* kernel. This closes that gap: the Yocto image from this branch was deployed as a CVM on a TDX host (`kvm_intel.tdx=Y`) through `dstack-vmm`, and the probe ran the guest's own binaries against the guest's own kernel.

Setup: `dstack-vmm` + supervisor, image installed from this branch's release tarball (`metadata.json` records `git_revision 51a577913`), app deployed with `--key-provider none`. The probe is a privileged `network_mode: host` container that mounts the guest root and `chroot`s into it, so every command below is the shipped `/usr/sbin` binary running in the CVM's network namespace on kernel `6.18.24-dstack`.

```
=====PROBE-BEGIN=====
guest kernel: 6.18.24-dstack
FRONTEND iptables -> /usr/sbin/xtables-nft-multi
  version: iptables v1.8.13 (nf_tables)
FRONTEND ip6tables -> /usr/sbin/xtables-nft-multi
  version: ip6tables v1.8.13 (nf_tables)
FRONTEND ebtables -> /usr/sbin/xtables-nft-multi
  version: ebtables v1.8.13 (nf_tables)

--F3 ebtables filter table--
Bridge table: filter
Bridge chain: INPUT, entries: 0, policy: ACCEPT

--F2 ip6tables nat--
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination

--F1 CHECKSUM (exact issue command)--
CHECKSUM exit=0
CHECKSUM   udp  --  0.0.0.0/0   0.0.0.0/0   udp dpt:68 CHECKSUM fill

--nft bridge family (Incus nftables driver)--
bridge table OK
bridge chain OK

CHECKSUM rules visible in nft ip mangle: 1
docker rules in nft: 33
=====PROBE-END=====
```

All three failures from #1032 are gone on the real image:

1. `-j CHECKSUM --checksum-fill` returns 0 and the rule is installed — previously `Extension CHECKSUM revision 0 not supported, missing kernel module?`
2. `ip6tables -t nat -L` lists the table — previously `Failed to list IPv6 rules (table nat)`
3. `ebtables -L` returns the filter table — previously `Module ebtables not found` / `the kernel doesn't support the ebtables 'filter' table`

Two results matter beyond the three failures:

**The nftables `bridge` family works at runtime.** Creating `table bridge incus` and a filter-hook chain in it both succeed. That is the path Incus's nftables driver renders per-instance bridge filters into, and it is what this PR chose over the legacy ebtables tables.

**There is genuinely one ruleset.** `docker rules in nft: 33` — dockerd's own rules are in nftables, not in a legacy table, so a container manager probing for the ruleset in use will find nftables and select its nftables driver. That is the mechanism behind the original report, now inverted. The CHECKSUM rule programmed through the `iptables` frontend is visible in `nft list table ip mangle` in the same ruleset.

### What is still not covered

Incus itself was not installed and run — the probe exercises the capabilities its managed-bridge paths depend on, not Incus's own code. The reporter's offer in #1032 to rerun their bridge and two-container probes on a disposable CVM is still the last mile, and this image should now satisfy it.

The mkosi image was not deployed as a CVM; its equivalent checks were done against the shipped artifacts and its own extracted kernel config, reported above.

kvinwang:
## Incus run for real, in the guest OS

Previous comment ran Incus inside a Docker container, which is not the reported scenario and hit cgroup-nesting limits unrelated to netfilter. This run installs Incus into the CVM's guest OS through `pre_launch_script` (`/usr` and `/etc` are tmpfs overlays, so this works on the read-only rootfs) and runs `incusd` in the guest's own root and network namespace — the reporter's setup.

mkosi image, kernel `6.18.40-dstack`, Incus 6.0.4.

### Incus's own nftables probe passes

Before anything else, `Nftables.Compat()` runs its functional check. The guest audit log caught it:

```
audit: table=incus_test_bca9aa08-...:27 family=2 op=nft_register_table   ... success=yes comm="nft"
audit: table=incus_test_bca9aa08-...:28 family=2 op=nft_unregister_table ... success=yes comm="nft"
```

That is the create-and-delete test table Incus uses to decide whether nftables really works rather than merely listing. It passes, so Incus selects its nftables driver — `table inet incus` appears in the ruleset — and never enters the xtables paths where all three reported failures live.

### A managed bridge with everything the reporter had to disable

```
Network dstacktest0 created
  ipv4.dhcp: "true"
  ipv6.address: fd99::1/64
status: Created
```

dnsmasq is running for it with the full DHCP + IPv6 RA configuration:

```
dnsmasq --interface=dstacktest0 --listen-address=10.77.0.1
  --dhcp-range 10.77.0.2,10.77.0.254,1h
  --listen-address=fd99::1 --enable-ra
  --dhcp-range ::,constructor:dstacktest0,ra-stateless,ra-names
```

The workaround in #1032 was to disable Incus DHCP and IPv6 and inject static addresses. Both are on here.

### Two system containers, addresses from DHCP and SLAAC

```
| NAME |  STATE  |        IPV4        |               IPV6               |   TYPE    |
| c1   | RUNNING | 10.77.0.128 (eth0) | fd99::1266:6aff:fe49:c775 (eth0) | CONTAINER |
| c2   | RUNNING | 10.77.0.8   (eth0) | fd99::1266:6aff:fe28:e457 (eth0) | CONTAINER |
```

`grep -iE "checksum|ebtables|ip6tables|Failed to list|Failed applying"` over the incusd log: **clean**.

### One finding worth recording: Docker's FORWARD policy

Container-to-container traffic across the managed bridge was 100% lost. That is **not** a missing capability, and the diagnosis is worth stating because anyone running Incus in a CVM will hit it.

`br_netfilter` is loaded with `net.bridge.bridge-nf-call-iptables = 1`, so frames bridged between two containers still traverse the forward hook. There, Docker's chain has `policy drop`:

```
table ip filter {
  chain FORWARD {
    type filter hook forward priority filter; policy drop;
```

while Incus's accept rules live in its own table:

```
table inet incus {
  chain fwd.dstacktest0 {
    type filter hook forward priority filter; policy accept;
    ip version 4 iifname "dstacktest0" accept
```

nftables base chains in different tables are evaluated independently at the same hook, so Incus's `accept` does not override Docker's `drop`. One rule proves the diagnosis and fixes it:

```
# iptables -w 5 -I DOCKER-USER -i dstacktest0 -o dstacktest0 -j ACCEPT
--- before: 3 packets transmitted, 0 received, 100% packet loss
--- after:  3 packets transmitted, 3 received, 0% packet loss, rtt 0.155/0.157/0.159 ms
--- IPv6:   3 packets transmitted, 3 received, 0% packet loss, rtt 0.149/0.181/0.234 ms
```

This is the long-standing Docker-versus-bridge-manager interaction, not something this PR introduces — the reporter's own environment notes already say "Docker owns an active ruleset with `FORWARD DROP`". It is now documented in `docs/guest-netfilter-capabilities.md`.

Worth noting what the convergence bought here: `nft list ruleset` shows Docker's drop and Incus's accepts side by side. While the two lived in separate rulesets, neither `iptables -L` nor `nft list ruleset` alone showed both halves, and this would have been considerably harder to pin down.

### Coverage

| | Result |
| --- | --- |
| Incus nftables compat probe | passes, nftables driver selected |
| Managed bridge, DHCP + IPv6 | created, dnsmasq serving both |
| Two system containers | running, DHCP + SLAAC addresses |
| Container-to-container v4/v6 | works, after the standard DOCKER-USER rule |
| Incus netfilter errors | none |

Not covered: the same run on the Yocto image (its equivalent netfilter checks were done in the earlier CVM test), and Incus's MAC/IP filtering and ACL features, which the reporter also said they do not use.

kvinwang:
## Yocto image: same Incus run, plus a behaviour change I need to flag

Incus 6.0.4 on the Yocto image from this branch, kernel `6.18.24-dstack`, running in the guest OS. Because that userspace is Yocto-built rather than Debian, the Debian Incus tree is unpacked into `/dstack/incus-root` and run under `chroot` — filesystem root only, so netns and pidns stay the guest's and every bridge and rule lands in the CVM itself.

```
iptables -> /usr/sbin/xtables-nft-multi : iptables v1.8.13 (nf_tables)
ebtables -> /usr/sbin/xtables-nft-multi : ebtables v1.8.13 (nf_tables)

Network dstacktest0 created
  ipv4.dhcp: "true"
  ipv6.address: fd99::1/64
status: Created

# bridge seen from the guest, not inside the chroot:
dstacktest0      DOWN           10.77.0.1/24 fd99::1/64

table ip nat / ip filter / ip6 nat / ip6 filter / inet incus
dnsmasq running

netfilter errors: (none - clean)
```

`table inet incus` again, so Incus selects its nftables driver here too, and the bridge comes up with DHCP and IPv6 — the two things #1032's workaround had to disable.

System containers did not start on this image: `incusd forkstart ... exit status 1`. That is the chroot, not the image — LXC needs its cgroup and `/run` layout at the real root. On mkosi, where the Debian tree could be installed into the guest's own overlay without a chroot, both containers ran with DHCP and SLAAC addresses. Not worth another harness round: the netfilter surface under test is identical and verified on both.

## The Docker FORWARD interaction is a behaviour change, not just a quirk

I reported earlier that container-to-container traffic needs one `DOCKER-USER` accept rule. I called it normal upstream behaviour. That is true, but incomplete, and the missing half matters for this PR.

Incus's **xtables** driver installs its accept with:

```go
d.iptablesPrepend(ipVersion, comment, "filter", "FORWARD", "-i", networkName, "-j", forwardType)
```

— `-I` into Docker's *own* `filter FORWARD` chain, ahead of the `policy drop`. Its **nftables** driver instead builds `table inet incus` with its own base chains. nftables evaluates base chains in different tables independently at the same hook, so that accept no longer pre-empts Docker's policy.

So: on the pre-0.6.1 Yocto image forwarding worked with no extra rule, and after this PR it needs one. That is a consequence of moving the image to the nftables dataplane, not of the kernel additions.

Two things bound how much this matters:

- On the old image Incus could not create a working managed bridge at all — that is what #1032 reports — so there is little for this to regress against. A deployment using the reported workaround (DHCP and IPv6 disabled) is the case that will notice.
- It is one rule per bridge, and it is what every nftables-frontend distribution already requires.

**I did not measure this directly.** The pre-PR comparison CVM could not get Incus running (its Yocto userspace needs the same chroot, which is where container start fails), so the claim rests on the Incus source above plus the measured nftables side. Calling it out rather than leaving it implied.

It is documented in `docs/guest-netfilter-capabilities.md`, and it belongs in the release notes for the image.

## What the pre-PR comparison did confirm

The baseline CVM (`81da8404`, verified as an ancestor of this branch with none of its changes) booted far enough to show:

```
kernel: 6.17.6-dstack
iptables -> xtables-legacy-multi; iptables v1.8.11 (legacy)
ebtables -> ; ebtables: command not found
```

Direct confirmation of the PR's premise: the shipped image programmed the legacy tables, and had no `ebtables` command at all.

kvinwang:
## Does this break existing Docker Compose apps? Measured: no

The obvious reviewer question for a change that moves the guest's netfilter dataplane. The same compose file — user-defined network, service-name DNS, a published port, outbound traffic — was deployed to two CVMs at once: one on this branch's Yocto image, one on the pre-PR image (`81da8404`, legacy dataplane).

| Check | This PR (nft, 6.18.24) | Pre-PR (legacy, 6.17.6) |
| --- | --- | --- |
| `curl http://web` — service DNS + user bridge | `http=200` | `http=200` |
| service name resolution | `172.18.0.2` | `172.18.0.2` |
| outbound internet (DNS + masquerade) | `http=200` | `http=200` |
| container → gateway | 0% loss | 0% loss |
| **published port, from outside the CVM** | **200** | **200** |

Identical, down to the address Docker hands out.

The reason is that Docker does not touch the kernel directly — it shells out to `iptables`. This PR only changes what that name resolves to. The rules Docker emits are unchanged in meaning; they land in nftables instead of the legacy tables, and Docker writes, reads and deletes them all through the same frontend. An earlier probe in a CVM counted 33 Docker rules in `nft list ruleset`, all functional.

### What *is* affected

Only a second thing in the CVM that manages its own bridges — Incus, LXD, libvirt. That is the FORWARD-policy behaviour change written up in the previous comment and in `docs/guest-netfilter-capabilities.md`: one `DOCKER-USER` accept rule per managed bridge. Plain compose apps, including ones that publish ports, are unaffected.

### Unrelated finding

The pre-PR guest agent could not parse the app-compose produced by the current `vmm-cli`:

```
Error: Failed to deserialize response
Caused by:
    EOF while parsing a value at line 1 column 0
```

The app still ran and served traffic, so it did not affect the comparison above. It is a separate version-compatibility issue between current `vmm-cli` and older guest agents, not something this PR touches.

PR #1054, opened 2026-08-14
feat(dstack-util): add streaming encrypt and decrypt commands

kvinwang:
## Problem

The CVM encrypted-environment path only decrypts one monolithic blob during setup. Applications that deliver encrypted data through another channel have no reusable guest-side decrypt command, and the legacy single-tag AES-GCM format requires buffering the complete message.

## Fix

- add `dstack-util encrypt`, which retrieves the app-scoped environment encryption public key from KMS
- require `--kms-pubkey` and verify the KMS timestamped `signature_v1` before encrypting
- add a magic-prefixed, versioned chunk format with independently authenticated AES-256-GCM frames
- add `dstack-util decrypt`, which auto-detects chunked ciphertext while retaining legacy encrypted-env compatibility
- support stdin/files, configurable chunk sizes, custom KMS root CAs, and legacy hex ciphertext input
- normalize KMS URLs with or without trailing `/prpc/`
- document the wire format and streaming failure semantics

Each frame authenticates the stream header, chunk index, flags, and plaintext length. The decoder rejects truncation, reordering, unknown flags, oversized frames, authentication failures, and trailing data. Because authenticated chunks are emitted incrementally, callers must discard output whenever the command exits unsuccessfully.

## Testing

- `cargo clippy -p dstack-util -- -D warnings -D clippy::expect_used -D clippy::unwrap_used --allow unused_variables`
- `cargo test -p dstack-util` (87 passed)
- round trips across multiple chunks
- tampered and truncated ciphertext rejection
- magic auto-detection and legacy fallback
- hex input decoding and URL normalization
- trusted signer recovery and untrusted signer rejection


kvinwang:
Addressed in `88bb69d48`.

- `encrypt` now requires `--kms-pubkey` and verifies the timestamped `signature_v1` against that trusted secp256k1 signer, including the 300-second default freshness window and 60-second future-skew allowance. It fails closed on missing, malformed, stale, or untrusted signatures.
- URL normalization now trims trailing slashes before detecting `/prpc`; this fixes both `encrypt` and `get-keys`.
- The format documentation now states that streaming output can contain an authenticated but incomplete prefix on failure and that callers must check the exit status and discard output.
- New error messages use lowercase initial letters.
- Added tests for magic auto-detection, legacy fallback, hex decoding, URL normalization, valid signer recovery, and untrusted signer rejection.

Verified with the repository strict clippy invocation and all 87 `dstack-util` tests.

PR #1055, opened 2026-08-14
vmm: return GPU reset errors from VM APIs

kvinwang:
## Summary

- flatten the full VM start error chain before crossing the pRPC boundary
- apply the same behavior to both `CreateVm` and `StartVm`
- add a regression test covering nested GPU reset errors

## Motivation

GPU sanitization added in #1048 can fail before QEMU starts. Some pRPC transports preserve only the outer error's `Display` value, causing the actionable PCIe reset failure to be hidden behind a generic VM start error. Flattening the chain at the VMM API boundary ensures clients receive the complete diagnostic.

## Tests

- `cargo fmt --all -- --check`
- `cargo test -p dstack-vmm` (117 passed)

### DIFF
--- a/dstack/vmm/src/main_service.rs
+++ b/dstack/vmm/src/main_service.rs
@@ -7,7 +7,7 @@ use std::ops::Deref;
 use std::path::{Component, Path, PathBuf};
 use std::time::{SystemTime, UNIX_EPOCH};
 
-use anyhow::{bail, Context, Result};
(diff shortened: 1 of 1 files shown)


kvinwang:
Closing after verification: the existing pRPC error encoder already serializes the full anyhow error chain, including low-level GPU reset I/O errors such as Permission denied. The proposed flattening was therefore redundant.

PR #1058, opened 2026-08-14
vmm: sanitize GPUs via VFIO hot reset instead of sysfs SBR

Leechael:
## Summary

- switch GPU sanitize-on-attach (#1048) from root-only sysfs writes to the `VFIO_DEVICE_PCI_HOT_RESET` ioctl so it works with the VMM running as an unprivileged user
- validate the kernel-reported reset scope: every device affected by the reset must belong to the GPU's own IOMMU group, otherwise the launch is aborted
- drop the `/sys/bus/pci/drivers_probe` re-probe; devices stay bound to vfio-pci across the reset
- add a `dstack-vmm sanitize-gpu <slot>...` subcommand for manually resetting GPUs through the exact launch-time code path

## Motivation

#1048 cannot be enabled in a common operations setup: `dstack-vmm` running as a dedicated unprivileged user without sudo. It issues the Secondary Bus Reset by writing Bridge Control in the upstream bridge's sysfs config space and re-probes devices through `/sys/bus/pci/drivers_probe`, and both files are writable by root only.

`VFIO_DEVICE_PCI_HOT_RESET` makes the kernel perform the same Secondary Bus Reset on the parent bridge, but it is authorized by device ownership instead of privilege: the caller presents an fd for every VFIO group affected by the reset. The `/dev/vfio` group nodes are the same ones QEMU opens to attach the GPU, so the VMM user already has access to them. This is also the mechanism QEMU itself uses to bus-reset devices without a usable FLR.

A single group fd suffices because every sanitized GPU sits alone behind a dedicated PCIe bridge and alone in its IOMMU group. This matches the typical passthrough host setup: H200-class GPUs behind dedicated downstream ports, one device per IOMMU group, vfio-pci bound at boot, and the VMM user granted `/dev/vfio` access via the same group membership QEMU requires. A read-only survey of production H200 hosts confirmed this topology and access model.

The privilege constraints and topology preconditions are documented at the top of `gpu_reset.rs`.

## Safety

- the dedicated-bridge topology check from #1048 is retained as defense
- `VFIO_DEVICE_PCI_GET_HOT_RESET_INFO` output is validated before resetting: any affected device outside the GPU's own IOMMU group aborts the launch, and the affected set is logged
- all VFIO fds are closed before QEMU is spawned; a VFIO group can be opened by only one process at a time
- the ioctl either performs the bus reset or fails; it cannot silently degrade to FLR

## Tests

- `cargo check -p dstack-vmm`; `cargo clippy` clean
- `cargo test -p dstack-vmm` (117 passed)
- **not yet validated on GPU hardware** — no test GPU is currently available. Planned validation once one is: a kprobe on `pci_bridge_secondary_bus_reset` to confirm the kernel executes the SBR on the expected bridge, config-space polling for the link-down window during reset, an A/B comparison against the sysfs method, and a reproduction of the original SPDM-timeout incident. `sanitize_on_attach` stays config-gated until then, and `dstack-vmm sanitize-gpu` runs the same path standalone for the experiment.

### DIFF
--- a/dstack/vmm/src/gpu_reset.rs
+++ b/dstack/vmm/src/gpu_reset.rs
@@ -2,9 +2,36 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+//! GPU sanitization before QEMU attach.
+//!
+//! Terminating QEMU while GSP/SPDM initialization is in progress can leave
+//! stale FSP/SPDM session state in a GPU. VFIO's attach-time FLR does not
+//! clear that state; a PCIe Secondary Bus Reset does. The next guest
+//! otherwise fails with an SPDM timeout followed by GSP/RmInitAdapter errors.
+//!
+//! The VMM runs as an unprivileged user, so it cannot issue the SBR by
+//! writing Bridge Control in the upstream bridge's sysfs config space: that
+//! file is writable by root only, as is /sys/bus/pci/drivers_probe. Instead
+//! the VFIO_DEVICE_PCI_HOT_RESET ioctl asks the kernel to perform the same
+//! bus reset. The ioctl is authorized by device ownership rather than
+//! privilege: the caller must present an fd for every VFIO group affected by
+//! the reset. The group nodes under /dev/vfio are the same ones QEMU opens
+//! to attach the GPU, so the VMM user already has access to them.
+//!
+//! A single group fd suffices only because every sanitized GPU sits alone
+//! behind a dedicated PCIe bridge and alone in its IOMMU group. The bridge
+//! topology is validated first, and the kernel-reported set of devices
+//! affected by the reset must all belong to the GPU's own group; anything
+//! else aborts the launch rather than risking disruption to other devices.
+//!
+//! A VFIO group can be opened by only one process at a time, so every fd is
+//! closed again before QEMU is spawned.
+
 use std::{
     collections::BTreeSet,
     fs::{File, OpenOptions},
+    mem::size_of,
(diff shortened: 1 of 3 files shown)


kvinwang:
Overall, switching from privileged sysfs writes to the VFIO hot-reset API is the right direction. I have two implementation suggestions:

1. The ioctl constant should follow the canonical Linux UAPI name: `VFIO_DEVICE_GET_PCI_HOT_RESET_INFO`. The current `VFIO_DEVICE_PCI_GET_HOT_RESET_INFO` changes the word order and makes it harder to compare the implementation with `<linux/vfio.h>`.

2. We should avoid maintaining handwritten VFIO ioctl numbers, C ABI structures, variable-length buffer parsing, and container/group/device lifecycle handling in dstack. The `rust-vmm/vfio` project already provides `vfio-bindings` and most of the required `vfio-ioctls` abstractions, but it currently does not expose the PCI hot-reset operations as public APIs.

We have created a temporary fork that adds the missing APIs:

- https://github.com/kvinwang/vfio/tree/feat/pci-hot-reset
- Commit: https://github.com/kvinwang/vfio/commit/d509db2

I suggest using this fork temporarily, validating the complete reset flow on the target GPU hardware, and then submitting the APIs upstream to `rust-vmm/vfio`. Once upstream publishes a release, dstack can switch back to the crates.io dependency.

This should substantially reduce the amount of unsafe VFIO ABI code maintained in dstack and give the hot-reset implementation a more appropriate reusable home.

PR #1063, opened 2026-08-17
feat(vmm): exec single-process launches in place

kvinwang:
## Summary

- add LaunchSpec::is_single_process to identify launches without swtpm
- install prepared file descriptors and exec QEMU in place for single-process launches
- keep the existing launcher orchestration path for QEMU plus swtpm

This keeps the process ID observed by Supervisor or systemd attached directly to QEMU and preserves the native QEMU exit status. Multi-process launches still need the launcher to coordinate startup, signals, shutdown, and cleanup.

## Stack

- based on #1061, which introduces the general launcher open-file path
- only the exec-in-place optimization is included in this PR

## Validation

- cargo clippy for dstack-vmm with all targets and warnings denied
- cargo test for dstack-vmm: 121 passed

### DIFF
--- a/dstack/vmm/src/app.rs
+++ b/dstack/vmm/src/app.rs
@@ -6,7 +6,8 @@ use crate::{
     config::{Config, NetworkFilterMode, Networking, NetworkingMode, ProcessAnnotation, Protocol},
     logrotate,
     netd::{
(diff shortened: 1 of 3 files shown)


kvinwang:
Closing this PR because I misunderstood the requested split. The exec-in-place work belongs in #1061; only the PrepareBridge RPC rename will be submitted separately.

PR #1066, opened 2026-08-17
docs(sdk): state that *Secure helpers only hash GetTlsKeyResponse

0xkurious:
The `*Secure` chain helpers (`toViemAccountSecure`, `toKeypairSecure`, `to_account_secure`, `to_keypair_secure`) documented that they SHA-256 the complete key material, but only the `GetTlsKeyResponse` branch does. For a `GetKeyResponse` they return the same key as their deprecated counterparts, so migrating from the deprecated helper to the `*Secure` one changes nothing on the `getKey()` path.

This PR corrects the documentation rather than the behavior:

- `*Secure` docstrings now state that a `GetTlsKeyResponse` is hashed and a `GetKeyResponse` key is used as is.
- The deprecated helpers' docstrings scope their security concern to the TLS-key path and note that a `GetKeyResponse` gives the same result as the `*Secure` variant.
- The Python and JS READMEs say the same.

Why not hash `GetKeyResponse` too: `getKey()` already returns a uniformly random KMS-derived key bound to its path and purpose, so another SHA-256 adds no cryptographic strength, while changing the behavior would move the address of every existing `getKey()` + `*Secure` caller. Domain separation is available at the KMS layer by deriving wallet keys from a dedicated path.

Reported by @0xkurious; the original implementation fix was replaced with this documentation fix, as offered in the original description.

### DIFF
--- a/sdk/js/src/__tests__/solana.test.ts
+++ b/sdk/js/src/__tests__/solana.test.ts
@@ -80,5 +80,18 @@ describe('solana support', () => {
 
       consoleSpy.mockRestore()
     })
+
+    it('should hash the full key material for GetKeyResponse (regression: previously fell through unhashed)', async () => {
+      const client = new DstackClient()
+      const result = await client.getKey('/', 'test')
+
+      const legacyKeypair = toKeypair(result)
+      const secureKeypair = toKeypairSecure(result)
+
+      // toKeypairSecure must apply SHA256 to the full key material, as its
+      // docstring promises. If it silently falls through to the raw key (the
+      // bug this test guards against), both public keys are identical.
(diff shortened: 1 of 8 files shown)


kvinwang:
Thanks for the careful report, @0xkurious. We went with the documentation option you offered: `getKey()` output is already a uniformly random KMS-derived key bound to its path and purpose, so an extra SHA-256 adds no strength, and changing the behavior would move the address of every existing `getKey()` + `*Secure` caller. I've force-pushed a docs-only commit to this branch that states plainly which branch is hashed, and scopes the deprecation warning to the TLS-key path. Your original commit is `870325a1a3` if you want it back.

PR #1078, opened 2026-08-20
feat(gateway): gate instances out of load balancing from the admin API

kvinwang:
## Problem

A CVM registers with the gateway from `dstack-util setup`, during boot, before
`app-compose.service` has even started the containers. From that moment the
instance is in `state.apps[app_id]` and eligible for traffic. The only filter
`select_top_n_hosts` applies is WireGuard handshake freshness
(`handshake_stale`, default 30m), which says the tunnel is up and nothing about
whether the application can serve a request.

That leaves an operator with no way to stop traffic to one instance of a
multi-instance app. Today the options are to shut the CVM down or to remove its
instance record — both of which destroy the thing you were trying to look at.
There is no way to say "stop sending this one new work, but leave it running so
I can find out what is wrong with it".

## Fix

Add a per-instance operator traffic gate, set through `Admin.SetInstanceReady`.

- **It only gates multi-target selection.** The gate is applied on the app-id
  path in `select_top_n_hosts` and `random_select_a_host`. Both functions
  already return early for a direct instance-id lookup, before any filtering,
  so `<instance_id>.<base_domain>` keeps resolving to a gated instance. That is
  the point: an instance pulled from rotation has to stay reachable to be
  investigated.

- **It does not fail open.** A health check is inference and may be wrong, so
  it should fail open; an operator instruction is not, so this one does not.
  Gating every instance of an app makes the app refuse new connections rather
  than quietly overriding the operator.

- **It survives re-registration.** Instance ids are derived from attestation
  and are stable across reboots, so a quarantined instance that reboots must
  not rejoin the rotation on its own.

- **It is carried in the instance record.** Every writer rewrites the whole
  record from its own in-memory copy, so a copy without the gate drops it. That
  happens two ways: a build predating the field reads the record fine (msgpack
  named maps) and just does not write the field back — one rolled-back node is
  enough, no cluster needed — and a node on the *current* build that has not
  seen the gate yet can clobber it via the lazy port-policy fetch, which
  rewrites the record on the first proxied connection. A separate key would
  survive both.

  It is still not worth one. `port_policy` and `admin_port_policy` ride in the
  same record, are rewritten by the same code, and are lost the same two ways,
  so guarding one field of three buys inconsistency rather than safety. The
  cost does differ and that is worth stating: a lost `port_policy` is
  re-fetched and a lost `admin_port_policy` falls back to what the instance
  reports, whereas a lost gate silently returns an instance to rotation.
  Re-issuing the gate rewrites the record and fixes it.

  Keeping it in `inst/` also means it reaches other nodes through the watcher
  that already exists, is removed with the instance, and cannot go corrupt
  independently of the record it describes.

  `InstanceData` is a msgpack named map, so a reader skips fields it does not
  know — widening it later stays a no-op for older nodes, which
  `an_instance_record_with_an_unknown_field_still_decodes` pins.

  If the record itself becomes unreadable, the instance is exempt from the
  removal pass and keeps what the data plane already holds, so the last known
  operator intent survives. Setting the gate again rewrites the whole record
  from memory, which repairs the unreadable one on the way past.

- **It takes effect immediately.** Setting the gate drops the app's `top_n`
  entry, which otherwise pins a pre-gate selection for up to `cache_top_n`
  (30s). An emergency cut-off that takes 30 seconds to apply is not one.

- **Existing connections are left alone.** They are already bridged; tearing
  them down would turn "stop sending it new work" into a second, louder
  outage.

Closing the last open instance of an app logs a warning once — draining an app
entirely is a legitimate thing to want, but silently refusing every connection
to it is worth saying out loud.

`Option<bool>` throughout, with `None` (nobody ever touched it) reading as
ready, so existing deployments and records are unaffected. `HostInfo` gains an
`optional bool admin_ready` and the dashboard grows a `Traffic` column, so an
instance left gated after an investigation is visible rather than forgotten.

## Verification

`cargo test -p dstack-gateway --all-features`: 180 passed, 0 failed.
`cargo fmt --all --check` and `cargo clippy -p dstack-gateway --all-targets -- -D warnings` are clean.

New tests cover the behaviours above individually, in
`main_service/tests.rs` unless noted:

- `gating_an_instance_removes_it_from_rotation_but_keeps_it_directly_reachable`
- `gating_an_instance_invalidates_the_selection_cache`
- `gating_every_instance_refuses_traffic_instead_of_failing_open`
- `the_gate_survives_re_registration`
- `the_random_selection_path_honours_the_gate` (the `connect_top_n = 0` path
  carries its own copy of the filter)
- `gating_an_unregistered_instance_is_an_error`
- `a_gate_arriving_through_kv_invalidates_the_cached_selection`
- `removing_an_instance_removes_its_gate`
- `an_empty_candidate_group_is_not_blamed_on_the_port_policy`
- `an_instance_record_with_an_unknown_field_still_decodes` (`kv/mod.rs`) — the
  gate rides in the instance record, so its forward compatibility is the
  record's: a build that adds a field must not make the record unreadable to
  one that does not know it.

## Notes

- This is the operator half of the traffic gate. A follow-up PR adds
  `healthy`, polled by the gateway from a new `Worker.Health` RPC on the guest
  agent, and the two compose: an instance takes traffic when it is both healthy
  and not gated.
- `admin_port_policy` is carried across the instance-rebuild path here too. It
  had the same hole and the fix is one line, and leaving two operator overrides
  behaving differently under a comment that argues for one of them would have
  been worse than fixing both.
- A failed write to the store is reported rather than swallowed, and the error
  says how far it got: the change is in force on this node only, is not
  durable, and will be undone by the next reload — not "until restart", since
  any peer writing an unrelated `inst/` record triggers a reload that reads the
  gate back out of the store.

### DIFF
--- a/dstack/gateway/src/kv/mod.rs
+++ b/dstack/gateway/src/kv/mod.rs
@@ -1179,18 +1179,33 @@ impl KvStore {
         let gate = self.delete_persistent(keys::admin_ready(instance_id), FailWrite::ADMIN);
         let override_policy =
             self.delete_persistent(keys::admin_port_policy(instance_id), FailWrite::ADMIN);
+        let observations = self.sync_forget_local_observations(instance_id);
+
+        let previous = previous?;
+        gate?;
+        override_policy?;
+        observations?;
+        Ok(previous.is_some_and(|entry| !entry.is_deleted()))
+    }
+
+    /// Withdraw this node's observations of an instance.
+    ///
+    /// `conn/` and `handshake/` are keyed by observer, so a delete can only
+    /// withdraw the deleting node's own -- they are the only ones it owns.
+    /// Every other node has to withdraw its own when it learns of the deletion,
+    /// or its observation outlives the CVM it describes: nothing sweeps those
+    /// prefixes, and a node's recycle loop cannot see an instance it has
+    /// already dropped from memory. They are ephemeral, so a restart collects
+    /// them, but a gateway's uptime is measured in months.
+    pub fn sync_forget_local_observations(&self, instance_id: &str) -> Result<()> {
         let conn = self.delete_ephemeral(keys::conn(instance_id, self.my_node_id), FailWrite::CONN);
         let handshake = self.delete_ephemeral(
             keys::handshake(instance_id, self.my_node_id),
             FailWrite::HANDSHAKE,
         );
-
-        let previous = previous?;
-        gate?;
-        override_policy?;
         conn?;
         handshake?;
-        Ok(previous.is_some_and(|entry| !entry.is_deleted()))
+        Ok(())
     }
 
     fn delete_persistent(&self, key: String, _which: u8) -> Result<Option<wavekv::types::Entry>> {
--- a/dstack/gateway/src/kv/mod.rs
+++ b/dstack/gateway/src/kv/mod.rs
@@ -214,9 +214,13 @@ pub struct LoadedOverrides {
     pub decoded: BTreeMap<String, AdminOverrides>,
     /// Which of an instance's override keys exist, readable or not.
     ///
-    /// Separate from `decoded` because absent and cleared are different
-    /// answers: a cleared port-policy override is a present record holding
-    /// `None`, and only this tells it from a key that was never written.
+    /// Carried beside `decoded` because absent and cleared are different
+    /// answers and `decoded` can only show one of them: a cleared port-policy
+    /// override is a present record holding `None`, which is the same `None` a
+    /// key that was never written produces. The two together are three states
+    /// -- no key, a key saying "cleared", a key carrying a value -- and every
+    /// reader needs all three, because only the first may fall back to the copy
+    /// an older build left in the instance record.
     pub present: BTreeMap<String, BTreeSet<String>>,
     /// Instance IDs whose override record is present but no longer decodes,
     /// mapped to the decode error.
@@ -228,6 +232,25 @@ pub struct LoadedOverrides {
     /// never gated, so the instance stays reachable for investigation either
     /// way, which is the property the gate itself is built on.
     pub unreadable: BTreeMap<String, String>,
+}
+
+impl LoadedOverrides {
+    /// Which override keys an instance actually has, readable or not.
+    ///
+    /// Answered per key because they are per key. A caller that asks per
+    /// instance gets the two confused, and confusing them drops an override.
+    pub fn written_keys(&self, instance_id: &str) -> WrittenKeys<'_> {
+        WrittenKeys(self.present.get(instance_id))
+    }
+}
+
+/// The override keys one instance has. See [`LoadedOverrides::written_keys`].
+pub struct WrittenKeys<'a>(Option<&'a BTreeSet<String>>);
+
+impl WrittenKeys<'_> {
+    pub fn contains(&self, name: &str) -> bool {
+        self.0.is_some_and(|names| names.contains(name))
+    }
 }
 
 /// The `inst/` records currently in the KV store, split by readability.
@@ -1099,8 +1122,8 @@ impl KvStore {
         let existing = self.load_all_instance_overrides();
         let mut moved = Vec::new();
         for (instance_id, overrides) in legacy {
-            let present = existing.present.get(&instance_id);
-            let has = |name: &str| present.is_some_and(|names| names.contains(name));
+            let written = existing.written_keys(&instance_id);
+            let has = |name: &str| written.contains(name);
             if let Some(ready) = overrides.ready {
                 if !has(keys::ADMIN_READY) {
                     self.record_migration(
--- a/dstack/gateway/src/kv/mod.rs
+++ b/dstack/gateway/src/kv/mod.rs
@@ -49,12 +49,7 @@ pub use https_client::{AppIdValidator, HttpsClientConfig};
 pub use sync_service::{fetch_peers_from_bootnode, PersistentWriteNotifier, WaveKvSyncService};
 use tracing::{error, warn};
 
-use std::{
-    collections::{BTreeMap, BTreeSet},
-    net::Ipv4Addr,
-    path::Path,
-    time::Duration,
-};
+use std::{collections::BTreeMap, net::Ipv4Addr, path::Path, time::Duration};
 
 use anyhow::{Context, Result};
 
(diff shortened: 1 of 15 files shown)


kvinwang:
Both findings are correct and both are fixed. Thanks — the first one directly contradicted a claim in the PR description.

### 1. Other gateways kept a stale selection — addressed in `322029771`

Confirmed by reading `reload_instances_from_kv_store`: it materialises the new `InstanceInfo` (including `admin_ready`) but only touched `top_n` in the narrow case where an app became empty through an `app_id` change. So every node except the one that took the RPC kept routing to the gated instance for up to `cache_top_n`.

Fixed by comparing the fields a cached selection is actually built from, via a new `InstanceInfo::routing_inputs_differ`, and dropping the entry for both the old and the new `app_id` when any of them changed. As you suggested, it covers more than `admin_ready`:

- `ip` — copied straight into the cached `AddressInfo`, so a changed IP left the cache pointing at the old address. Pre-existing bug of the same class.
- `app_id` — decides which cache entry the instance belongs to
- `public_key` — the key the handshake-freshness filter looks up, so it gates eligibility
- `admin_ready` — the gate itself

A record this node has not seen before also invalidates, since it is a new candidate.

New test `a_gate_arriving_through_kv_invalidates_the_cached_selection` builds the cache, writes a gated record straight into the KV store the way a peer's sync round would, runs a reload, and asserts both that `top_n` is empty and that the next selection excludes the gated instance.

### 2. Persistence failure reported as success — addressed in `9cd47ff68`

`persist_instance_record` now returns `Result` and `set_admin_ready` propagates it, with context spelling out the exact state the operator is in: *"instance X is now ready=false on this node only: the setting could not be persisted, so it will not reach the other gateways and will be lost when this one restarts"*.

On your explicit question about rolling back the in-memory change: **deliberately not rolled back.** The gate is already in force on this node, and for an emergency cut-off partial protection beats undoing it — what must not happen is the operator being told it succeeded everywhere. This is documented at the call site.

The three pre-existing callers (`set_admin_port_policy`, `clear_admin_port_policy`, `update_instance_port_policy`) keep logging-and-continuing, now explicitly with a comment saying so rather than by accident. They have the same weakness, but changing `SetInstancePortPolicy`'s error contract is a different API's semantics and belongs in its own change.

`cargo test -p dstack-gateway --all-features`: 181 passed. `cargo fmt --check` and `clippy -D warnings` clean.

kvinwang:
Reviewed all four. Three are correct and fixed in `dc06a3a91`; the fourth is half right and I've fixed the half that was.

### #4 — mixed-version rollout clears the gate: **correct, and worse than stated**

This is the important one. Confirmed the mechanism: `InstanceData` is encoded with `rmp_serde::to_vec_named` and decoded with plain serde, so a node whose struct lacks the field silently drops it on decode — and then rewrites the record without it.

The finding says "某 CVM 重启后向尚未升级的节点注册". That path is real, but it is not the widest one. Three production paths rewrite the whole `inst/<id>` record from the writer's own struct:

- `new_client_by_id` on re-registration (boot only)
- `add_instance`
- **`persist_instance_record`, reached from `update_instance_port_policy`** — the lazy port-policy fetch, which fires on the *first proxied connection* to any instance with no cached policy

So it does not need a reboot. An old node serving one connection to a gated instance is enough to drop the gate, and LWW then clears it cluster-wide.

Fixed by moving the gate out of the shared record into its own `admin_ready/<instance_id>` key, following the existing `conn/` and `handshake/` per-instance key precedent. A build with no code for that key cannot clobber it. `sync_delete_instance` now deletes it alongside the conn and handshake records.

Two regression tests: `rewriting_the_instance_record_does_not_clear_the_gate` (writes the full record the way any node would, reloads, asserts the gate survives and selection still excludes the instance) and `removing_an_instance_removes_its_gate`.

Worth noting the same class of bug applies to `port_policy` / `port_policy_hash` / `admin_port_policy`, which predate this PR. Those degrade to a re-fetch or a revert-to-instance-reported rather than to silently un-quarantining a box, so I have not moved them — but the pattern is now available if you want them moved later.

### #1 — `admin_port_policy: None` on the IP-rebuild path: **correct**

I had seen this and left it as pre-existing, and said so in the PR description. That was the wrong call for the reason you give: the diff adds a comment arguing why `admin_ready` must survive that rebuild, and the identical argument applies to the override sitting on the next line. Leaving two operator overrides behaving differently under a comment that justifies one of them is worse than fixing both. Now carried across as well.

### #3 — port-policy error masks an operator drain: **correct**

Verified the ordering: `filter_allowed_addresses` runs before `connect_multiple_hosts`, so an empty group from selection produces `port 443 denied by app port policy for X (0 candidate(s))` and the clearer "no addresses to connect to app" never runs. `filter_allowed_addresses` now bails early when it was handed nothing, with `no instance of app X is eligible to serve port N` — nothing was offered, so nothing was denied here.

### #2 — `ready=true` persistence failure: **half right**

The behaviour and the error text were already correct for both directions. The message is *"instance X is now ready=true on this node only: the setting could not be persisted, so it will not reach the other gateways and will be lost when this one restarts"* — it says "on this node only", not "not restored", so an operator is told what actually happened.

What was wrong is the **inline comment**, which only argued the `ready=false` case ("for an emergency cut-off partial protection beats undoing it"). Rewritten to cover both: closing the gate is emergency containment, opening it is emergency recovery, and half of what the operator asked for beats none of it in either direction — what must not happen is being told it took effect everywhere.

I'm keeping no-rollback. Rolling back on a failed write would mean a failed cut-off leaves traffic flowing and a failed restore leaves the node gated, in both cases undoing protection the operator asked for to preserve a cluster-consistency property the error message already reports as broken.

---

`cargo test -p dstack-gateway --all-features`: 183 passed. `cargo fmt --check` and `clippy -D warnings` clean.

dstack/gateway/src/main_service.rs
```diff
@@ -1264,7 +1293,8 @@ impl ProxyState {
             reg_time: SystemTime::now(),
             port_policy,
             port_policy_hash: compose_hash.to_string(),
-            admin_port_policy: None,
+            admin_port_policy: carried_admin_port_policy,
+            admin_ready: carried_admin_ready,
```
kvinwang: **These two carried fields are the one thing the PR summary calls out ("survives re-registration"), and they have zero coverage.**

`carried_admin_ready` / `carried_admin_port_policy` only matter on the **rebuild** branch — the one taken when `!self.valid_ip(existing.ip)` and a fresh `InstanceInfo` is constructed. `the_gate_survives_re_registration` (`tests.rs:555`) re-registers with the same id/app/key and a *valid* IP, so it takes the early-return path above where the in-memory record is never rebuilt and these locals are unused.

Mutation-verified: replacing `admin_port_policy: carried_admin_port_policy, admin_ready: carried_admin_ready` with `None, None` leaves **all 180 gateway tests passing**, including that one. `grep` for `client_ip_range` / `valid_ip` in `tests.rs` returns nothing — no test exercises the invalid-IP rebuild path at all.

Second, smaller point on the same lines: `admin_port_policy` previously became `None` on this branch, dropping the operator's port override and falling back to the instance-reported policy. It now survives. That is plausibly the right fix, but it is a change to a *port-authorization* override shipped inside a "traffic gate" PR, it is not in the CHANGELOG, and it is untested for the same reason.

dstack/gateway/src/main_service.rs
```diff
@@ -1336,15 +1374,83 @@ impl ProxyState {
...
+        // here would just reach the same end state sooner while losing the brief
+        // window in which the operator's intent is at least locally in force.
+        self.persist_instance_record(instance_id).with_context(|| {
+            format!(
+                "instance {instance_id} is set to ready={ready} on this node \
+                 only and not durably: the write to the store failed, so it \
+                 will not reach the other gateways and will be undone here by \
+                 the next reload. Retry before relying on it"
+            )
+        })?;
```
kvinwang: **The error text promises something that does not happen in a quiet cluster, and the contract it exists to enforce is untested.**

(a) The message says the change "will be undone here by the next reload", and the comment above says "seconds, not until restart". But reload is only triggered by `watch_instances` → `watch_prefix(INST_PREFIX)` (`kv/mod.rs:1018`, consumed at `main_service.rs:973-982`). In a cluster where no instance registers, re-registers, or is deleted, no `inst/` key changes and **no reload happens**. The node then enforces a gate indefinitely that no peer knows about and that no restart-surviving record contains — the opposite of what the operator was just told.

There is also a path in the other direction: the CVM's own 180s registration refresh rewrites the record through `new_client_by_id` → `sync_instance`, which can *persist* the gate the write failed to store. Either way, predicting the outcome is not something this message can do; better to state only what is known (not durable, not replicated, retry).

(b) Swallowing the `?` here — i.e. reverting to the old log-and-continue — leaves **all 180 tests passing**. The whole point of changing `persist_instance_record` to return `Result<()>` (`:1451`) was so this one caller could surface durability failures, and nothing pins it.

dstack/gateway/src/main_service.rs
```diff
@@ -1336,15 +1374,83 @@ impl ProxyState {
...
+            .filter(|info| info.is_admin_ready())
+            .count()
+    }
+
     /// Persist the current in-memory `InstanceInfo` snapshot to WaveKV.
-    fn persist_instance_record(&self, instance_id: &str) {
+    ///
+    /// Returns the store error rather than only logging it, so callers whose
+    /// API promises the change is durable can say otherwise when it is not.
+    fn persist_instance_record(&self, instance_id: &str) -> Result<()> {
```
kvinwang, line 1780: **Blind-put of the in-memory snapshot is how a peer silently reopens the gate cluster-wide.**

`persist_instance_record` rewrites the *whole* `inst/<id>` record from this node's `InstanceInfo`, including `admin_ready`. Combined with `update_instance_port_policy` (reached from `proxy/port_policy.rs:217-232`, which drops the lock for a network `Info()` RPC and then re-takes it):

1. node A takes `SetInstanceReady(X, false)` and writes `inst/X` with `admin_ready = Some(false)`;
2. before B's `watch_instances` reload delivers it, B rewrites `inst/X` from its own memory with `admin_ready: None`;
3. B's write is later, LWW makes it the winner, and A's next reload puts X straight back into rotation.

Nothing logs, warns, or increments a metric — the operator's `SetInstanceReady` returned `Ok`.

`InstanceData::admin_ready`'s doc acknowledges this and argues it is acceptable because `port_policy` / `admin_port_policy` lose the same way — but the doc itself concedes those two are self-healing and the gate is not. Minimum viable mitigation is to have this read-modify-write the stored record rather than blind-put the snapshot for fields it did not intend to change.

dstack/gateway/src/kv/mod.rs
```diff
@@ -102,6 +102,34 @@ pub struct InstanceData {
...
+    /// writer rewrites the whole record from its own in-memory copy, so a copy
+    /// that does not carry the gate drops it. Two paths get there:
+    ///
+    /// - A build that predates the field. Records are msgpack named maps, so it
+    ///   reads the record fine and just does not write the field back. This
+    ///   needs no cluster -- a single node rolled back is enough.
+    /// - A node on the *current* build that has not seen the gate yet. The lazy
+    ///   port-policy fetch rewrites the record on the first proxied connection,
+    ///   so between `SetInstanceReady` landing on one node and the sync reaching
+    ///   another, that other node can clobber it. Normally the window is one
```
kvinwang: **The conclusion holds, but this names the wrong path — and the real one is broader and far more frequent.**

The lazy port-policy fetch only fires when `instance_port_policy()` returns `None` (`proxy/port_policy.rs:68-81`), which for a current-build CVM never happens: it reports `port_policy` at registration (`dstack-util/src/system_setup.rs:486`). The exception is after a `compose_hash` change invalidates the cache — and *that* connection is itself denied (`DenyReason::PolicyUnknown`), so it is not a proxied connection either.

What actually clobbers the record is the CVM's **180-second registration refresh**, which rewrites the whole record through `new_client_by_id` → `sync_instance`. Same conclusion, much wider window, and it happens on a timer rather than on traffic.

The PR description carries the same wording and needs the same correction.

dstack/gateway/src/kv/mod.rs
```diff
@@ -739,14 +767,26 @@ impl KvStore {
...
+        let handshake = self
+            .ephemeral
             .write()
-            .delete(keys::handshake(instance_id, self.my_node_id))?;
+            .delete(keys::handshake(instance_id, self.my_node_id));
+
+        let previous = previous?;
+        conn?;
+        handshake?;
         Ok(previous.is_some_and(|entry| !entry.is_deleted()))
```
kvinwang: **This rewrite is entirely untested.**

The commit reworks this to attempt all three deletes even when an earlier one fails, and to report the `inst/` failure ahead of the ephemeral ones — with a 7-line comment justifying it. Reverting the body to the original `?`-short-circuiting version leaves **all 180 tests passing**.

`removing_an_instance_removes_its_gate` only pins that the three keys are deleted on the happy path, which the old code also did (it *is* non-vacuous for that: stubbing out the handshake delete makes it fail). Nothing injects a store failure, so the two properties this commit actually adds — no short-circuit, and the `inst/` error reported first — have no coverage.

kvinwang (commented):
Self-review of this PR. The gating logic itself came out clean under mutation testing, so these are precision and coverage findings rather than logic bugs — but three of them are on lines the PR description specifically calls out.

Verified clean (mutation-tested in a throwaway worktree at `c2a4ccf3b4`): both selection paths are gated and both gates are load-bearing (deleting the `is_admin_ready()` check in `select_top_n_hosts` fails 6 tests; neutering it in `random_select_a_host` fails `the_random_selection_path_honours_the_gate`); the instance-id path genuinely returns before the gate on both; `connect_top_n == 0`, the `latest_handshakes` error fallback and "all instances gated" all fail closed; every `top_n` invalidation site is load-bearing; no lock is held across network I/O outside the pre-existing lazy fetch. `180 passed`, `cache_top_n = "30s"`, `handshake_stale = "30m"` and all ten named tests in the description check out.

The five inline comments below:

1. `main_service.rs` — the carried `admin_ready` / `admin_port_policy` (the "survives re-registration" line) is on a branch no test reaches; replacing both with `None, None` keeps all 180 tests green.
2. `main_service.rs` — `set_admin_ready`'s error text predicts a reload that need not happen in a quiet cluster, and the error contract itself is untested.
3. `main_service.rs` — `persist_instance_record` blind-puts the in-memory snapshot, which is the mechanism by which a peer silently reopens the gate cluster-wide.
4. `kv/mod.rs` — the `admin_ready` doc (and the PR description) names the lazy port-policy fetch as the clobber path; the real one is the 180s registration refresh, which is broader and more frequent.
5. `kv/mod.rs` — the `sync_delete_instance` rewrite has no test for either property it adds.

PR #1079, opened 2026-08-20
feat(gateway): keep unhealthy instances out of load balancing

kvinwang:
> Stacked on #1078. Supersedes #1080 and #1082, both now closed — the whole
> health story is one PR.
>
> Revised after self-review ([inline comments](https://github.com/Dstack-TEE/dstack/pull/1079#pullrequestreview-4988832474)).
> The feature is now **opt-in**, the app-declared program probe is gone, and
> `Worker.Health` costs a lock and a clone. See *What changed after review* at
> the bottom for the full list.

## Problem

A CVM registers with the gateway from `dstack-util setup`, during boot.
`app-compose.service` is ordered *after* `dstack-prepare.service` and runs
`docker compose up -d`, so at the moment of registration the app's containers do
not exist yet — the image may still be pulling. From that instant the instance is
in `state.apps[app_id]` and eligible for traffic.

The only filter `select_top_n_hosts` applies is WireGuard handshake freshness
(`handshake_stale`, default 30m). That says the tunnel is up. It says nothing
about whether anything is listening behind it.

Racing the connect across `connect_top_n` candidates hides the easy half — a
container that is not listening yet refuses the TCP connection and loses the
race. It does not hide the common half: a container that has bound its port but
is still warming up accepts instantly (it has no load), *wins* the race, and then
fails at the application layer. There is no re-dispatch after a successful
connect, so the client eats the error.

## Fix

The gateway asks each opted-in guest agent whether the app is actually serving,
and keeps instances that say no out of app-id load balancing.

**Opt-in, under `requirements`.**

```json
"requirements": {
  "health_check": { "enabled": true, "health_file": "/dstack/health" }
}
```

Not a top-level field. `Requirements` is `deny_unknown_fields` and is itself
rejected below `manifest_version` 3, so a deployment that asks for gating cannot
land on a guest image that would silently ignore it — the same "so older guests
fail closed" rule `requirements`, `nerdctl-compose` and multi-`init_script`
already follow. An app that does not opt in is never polled and is routed to
exactly as before.

`enabled` is what reaches the gateway as `RegisterCvmRequest.health_check`.
Keeping it explicit rather than implied by the object's presence means turning
gating off during an incident is a one-word edit, and the compose hash records
the intent either way.

**Poll, don't push.** A wedged, OOM-killed or mid-crash guest agent shows up as a
failed poll, rather than as silence that has to be told apart from "nothing
changed" with a timestamp and a staleness heuristic. It keeps health a *per-node
observation*, exactly like the WireGuard handshakes it sits next to, instead of
shared state that outlives whatever wrote it. And it reuses the path
`proxy/port_policy.rs` already uses — plain HTTP over the WireGuard tunnel, no
RA-TLS client in the guest agent, no new outbound direction.

**New `Worker.Health` on the guest agent, and it is nearly free.** The agent
recomputes its verdict on its own 5s timer and caches it; the RPC takes a read
lock and clones. That is not an optimisation, it is the security property:
`Worker` is served on the CVM's public `0.0.0.0:8090` listener, so anything the
handler does per request is work an anonymous caller can ask for at an arbitrary
rate — and every gateway node in the cluster polls the same instance, so
per-request work also multiplies by the operator's fleet size.

**Exposure is unchanged.** `Worker.Health` sits on the same `agent_port` (8090)
surface `Worker.Info` already does, behind the same boundary: WireGuard is
hub-and-spoke here — the gateway's config carries one peer per CVM with
`AllowedIPs = <cvm_ip>/32`, while a CVM's config only lists gateway IPs — so a
CVM cannot route to another CVM.

**Capability declared at registration, not probed.** prpc answers both "no such
method" and "the handler failed" with HTTP 400, and `PrpcClient` drops the body,
so a failed poll cannot distinguish an old image from a broken app. The gateway
counts every failed poll as unhealthy, so probing to discover support would
blackhole exactly the instances that cannot answer. For the same reason
`Worker.Health` never returns an RPC error: a failure to see the app comes back
as `healthy: false` with `error` set.

**Health fails open; the operator gate does not.** If health would empty an app's
candidate set it is ignored and the app is routed to anyway — blackholing an app
on the strength of an inference is worse than sending traffic to instances that
might be fine. `admin_ready` from #1078 deliberately does not get this treatment:
that one is an instruction.

Worth being explicit about the consequence, since it is easy to over-read what
this feature buys: for a **single-instance app**, or a fleet that boots together,
fail-open routes traffic while every instance is still `Unknown`. Health gating
shifts traffic *away from a bad instance*; it does not hold an app offline until
a good one exists.

**Not-yet-answered is not healthy.** An instance that opted in starts `Unknown`
and stays out of rotation until a poll answers. A *reboot* resets it — detected
by a fresh WireGuard key, since the guest caches its key store in `/run`, which
is tmpfs — while the gateway-checker's 180s re-registration does not.

### Where a verdict comes from

The app picks one of two sources.

**`health_file`** — the app writes two lines, and the agent reads them:

```text
healthy
1771234567
```

State (`healthy`/`unhealthy`, case-insensitive) and the unix time it was
written. The timestamp is not decoration: a file older than 60s counts as
unhealthy, which is what turns a wedged app — still running, no longer updating
anything — into a verdict rather than a stale `healthy` that never expires.
Anything unreadable, malformed or unrecognised is also not healthy.

Reading a file replaces running a program. It cannot hang, cannot leak a
process, and needs no timeout, args, interval or kill semantics to get right.

**Container healthchecks (the default)** — `docker-compose` over the Docker
socket, `nerdctl-compose` through `nerdctl inspect`. Only containers that declare
a Compose `healthcheck` are judged; both runtimes populate health state
exclusively for those, so its absence is the app saying "I have not told you how
to test this one" rather than a pass — which keeps a one-shot init container that
exited cleanly from failing the instance. `starting` counts as not healthy.
Finding **no containers at all** is not healthy either.

A container that declares a healthcheck and is **not running** is not healthy
either, whatever verdict the runtime last recorded. Docker re-marks a stopped
container `unhealthy`; nerdctl leaves the last result in place, so a container
that passed one check and then crashed still reports `healthy` with
`Running: false` — and nerdctl is what the guest image runs.

Only the app's **own Compose project** is judged, matched on
`com.docker.compose.project`. Container state lives on the persistent data disk
(`dstack-prepare.sh` rbind-mounts `/var/lib/{docker,containerd,nerdctl}`), so it
outlives reboots and upgrades, and `--remove-orphans` only ever touches the
project it was invoked for. Without the filter, one exited container left behind
by an earlier deployment under a different project name would hold the instance
out of rotation permanently.

Measured, not assumed: nerdctl 2.3.5 sets exactly `com.docker.compose.project`,
`.service` and `.config-hash` — the docker-only `project.working_dir` label would
have matched nothing under the runner the guest image actually uses. Both
runtimes normalise the project name the same way (`name: MyApp_X` → `myapp_x`).

Runners with no containers to inspect (`bash`) are rejected at setup unless they
declare a `health_file`, rather than silently reporting healthy forever.

### nerdctl

nerdctl generates its inspect output from its own `dockercompat` types, so
`State.Health.Status` carries the *same four strings* Docker uses — the decision
rule is genuinely shared rather than reimplemented.

Compose `healthcheck:` needs **nerdctl >= 2.3.1**; from the tagged sources,
`pkg/composer/serviceparser` does not handle the field in 2.2.1 or 2.3.0 and does
from 2.3.1 on. `versions.env` moves the mkosi backend to 2.3.5; Yocto is already
there on `next` (`meta-virtualization` at `fdaa0d2de8`, `PV = "v2.3.5"`).
Container *presence* is detected on older versions regardless.

`compose_hash` is `sha256_file(app-compose.json)` — over raw file bytes, not a
re-serialisation — so the new optional field changes no existing app's hash.

## Configuration

```toml
[core.proxy.health_check]
enabled = true
interval = "5s"
timeout = "2s"
concurrency = 16
failure_threshold = 2
state_file = "/dstack-gateway/data/instance-health.json"
```

Default-on with a bounded blast radius, because the gate now lives in the app's
manifest: nothing is polled unless its app asked to be. `enabled = false` skips
the filter outright — not merely leaving it unfed, which would strand every
opted-in instance in `Unknown`.

`failure_threshold` is how many consecutive *unreachable* polls demote an
instance. An agent that answers "unhealthy" is believed on the spot: it can see
the app and the gateway cannot.

`state_file` is where a node remembers its own verdicts across a restart of the
process. It is stamped with the kernel's boot id and discarded when that changes
— nothing this node believed before the machine went down is evidence about what
is running after it — and each entry carries the WireGuard key it was observed
against, so a CVM that rebooted while the gateway was gone does not inherit a
verdict about a process that no longer exists.

5s is not the bottleneck: end-to-end detection is dominated by the app's own
`healthcheck` settings, where Docker's defaults (30s interval, 3 retries) take up
to 90s to mark a container unhealthy.

## Documentation

`docs/app-health-checks.md` covers #1078 + this PR together: how an app opts in,
where the verdict comes from, what container healthchecks do and do not judge,
the `health_file` contract, the nerdctl requirement, how the gateway uses the
result, why health fails open while the operator gate does not (and what that
means for a single-instance app), and the gateway-side config.

## Verification

```
dstack-gateway      223 passed
dstack-guest-agent   67 passed
dstack-util          95 passed
dstack-types         39 passed
http-client           9 passed
```

Plus new Go and Python SDK tests pinning that `requirements.health_check`
reaches the compose hash.

`cargo fmt --check` and `clippy -D warnings --all-targets` clean across gateway,
guest-agent, dstack-util and dstack-types.

Every behavioural claim above that a test is supposed to pin was checked by
mutation — deleting the code and confirming the named test fails. That includes
the ones this round added (stale-verdict guard, handshake filter, capability
inheritance, response-body cap, setup-time validation) and the one it fixed:
`disabling_polling_also_keeps_the_random_path_open` was **vacuous** — with two
`Unknown` instances the fail-open branch kept both regardless, so deleting the
check it was written to pin left the whole suite green. It uses a mixed fleet now
and fails on that mutation.

**Verified on hardware** (tdx-lab) earlier in this PR's life, and still valid:

*nerdctl 2.3.5 really does translate the healthcheck* —
`--health-cmd=true --health-interval=2s --health-timeout=2s --health-retries=2
--health-start-period=1s` in the run log. It also warns `CMD exec form is not
supported, converting to CMD-SHELL`, a real behavioural difference from Docker.

*The checks run, via systemd timers, and inspect reports the result* — a
three-service project gave `healthy` / no `Health` object / `unhealthy`, with
exactly two timers, one per container declaring a check. Two assumptions were
corrected and are now pinned by `real_nerdctl_inspect_output_parses` against the
captured payload: `Name` has **no** leading slash, and a container with no
healthcheck has **no `Health` object at all** rather than `"Status": "none"`.

*And it works inside a real TDX CVM with a read-only rootfs* — the open question
was whether a transient systemd unit has anywhere to land when `/` is immutable.
It does: the timer is transient and lands in `/run` (tmpfs), while nerdctl's
state is writable because `dstack-volatile-binds.sh` overlays `/var/lib` and
`dstack-prepare.sh` binds the encrypted data disk over `/var/lib/nerdctl`.

*Runtime behaviour used to write the collector* was measured on this host against
nerdctl 2.3.5 and docker 29.5.3 — including the one that decides the code shape:
a batch `nerdctl inspect` containing a single stale id exits 1 with **empty
stdout**, wherever the bad id sits, while docker still returns the entries it
found. Hence the per-id fallback.

## What changed after review

Self-review found 23 issues; an adversarial pass over the fixes found 10 more,
including one regression the first round introduced. The seven original
blocking ones and the reasons they are gone:

| # | Problem | Now |
|---|---|---|
| 1 | A timed-out probe was never killed — `tokio::process` does not `kill_on_drop`, so one root process leaked per interval, unbounded | Program probe removed entirely; `health_file` cannot hang. The nerdctl calls that remain set `kill_on_drop(true)` and carry a timeout |
| 2 | `Worker.Health` did unbounded work per anonymous request on the public listener | Answers from a cached verdict; all work is on the agent's own timer |
| 3 | No timeout anywhere in `collect()` | Per-call timeout plus an outer refresh timeout |
| 4 | One container vanishing mid-collection failed the whole report | Docker 404s skipped; nerdctl falls back to per-id inspect (its batch returns *empty* stdout on partial failure) |
| 5 | Judged set was "any compose project" / "the whole namespace", and container state survives reboots → possible permanent unhealthy | Judged set narrowed to the app's own project |
| 6 | Unbounded response body from a guest-controlled peer, read every 5s | The **caller** can now bound it (`PrpcClient::with_max_response_bytes`), and the two gateway→guest-agent clients do. Deliberately not a transport-wide cap: `dstack vmm logs --lines 100000` is a legitimate multi-megabyte fetch and a global limit would break it |
| 7 | The round was applied as one batch with no generation guard, so a pre-reboot verdict could land on a post-reboot instance | Applied as each poll lands, against the WireGuard key and IP that were polled |

And the rest: poll targets are filtered by handshake freshness (a few powered-off
CVMs could stretch a round past its interval and slow detection for the whole
fleet); a transient failure no longer ejects an instance; the poller task is
supervised rather than a bare `tokio::spawn` that could die silently; the
capability is `Option<bool>` in the KV record and in `ReportedCapabilities`, so
an older peer's rewrite — or a debug registration — no longer reads as an
opt-out; `warn_all_unhealthy` is rate-limited where it is emitted rather than
where its doc comment claimed; app-controlled bytes are truncated and stripped of
control characters before reaching a log; `health_check.rs` has tests.

### Found by reviewing the fixes

- **A transport-wide 1 MiB cap broke `dstack vmm logs`.** Same helper, and
  `--lines` is unbounded. The cap is now a caller's choice; unbounded callers
  are unchanged, with a test pinning that.
- **The generation guard had a slower twin.** The reboot reset is applied by
  whichever node took the re-registration; every *other* node learned about the
  new boot through KV reload, which carried its own pre-reboot verdict across
  on the capability alone. So the new guard closed a 2s window while a
  whole-interval one stayed open on every peer. The reload compares the
  WireGuard key now.
- **The snapshot's default path does not exist in the shipped gateway CVM.**
  `entrypoint.sh` generates the deployed config and emitted no
  `[core.proxy.health_check]` at all, so a new module would have been silently
  inert exactly where it was needed. It now points at the data volume, and
  `write_snapshot` creates its parent.
- **The gateway trusted the agent's sanitizing.** The agent strips control
  characters — but the agent is the untrusted party here. Re-bounded and
  stripped where the log line is emitted.
- **The SDKs dropped the new field from the compose hash**, which is what gets
  whitelisted on chain. Fixed in Go, Python and JS with tests. (The same closed
  shape still drops `requirements.gpu_policy` — pre-existing, left alone.)
- Plus: `compose_project` was re-parsing the compose YAML every 5s in a
  synchronous, exponentially-expandable parser; a `.max(1)` clamp was dead code
  with a test that could not detect its removal; two tests asserted a constant
  against itself; the guest agent warned once per interval forever once a
  runtime was down; and the persistence wiring had no test at all — deleting
  either the read or the write side left the suite green.

Three claims in the previous description were also wrong and are corrected above:
registration does **not** only run at boot (`cb7c8d8a96` on this branch fixed the
code, the text lagged); Yocto is **already** on nerdctl 2.3.5, so that "known
limitation" is gone; and `HostInfo` field 5 was `uint32 port`, removed by
`b325c0f8bf` when multi-port listening landed — squarely inside this repo's
visible history, not before it. Reserving it is still right; the stated reason
was not.

## Known limitations

- **Every poll builds a fresh HTTP client.** `http-client`'s
  `http_request_with_headers` does `reqwest::Client::builder().build()` per
  request, so nothing is pooled and each poll opens a new TCP connection.
  `proxy/port_policy.rs` already lives with this, but its fetch is lazy and
  once-per-instance, whereas polling makes it steady-state. The fix is a shared
  client inside `http-client`, which every component in the tree uses, so it
  belongs in its own change. `interval` and `concurrency` bound it meanwhile.
- **Polling cost is `nodes × opted-in instances`.** Opt-in cuts this a long way
  from where it started; the interval is the remaining knob.
- **`health_file` staleness is a fixed 60s.** Configurable would be one more
  field, and an app asking for a longer bound is asking for its own failures to
  take longer to notice. Easy to add later if a real workload needs it.


kvinwang:
Both findings are correct — the first one meant the PR did not actually fix the bug it was written for. Both fixed, plus a third of the same family that the review did not reach.

Rebased onto the updated #1078.

### 1. Empty container list read as healthy — addressed in `6b1f91d38`

Correct, and this was the whole bug walking back in. Registration happens in `dstack-prepare.service`; `app-compose.service` is ordered after it. At poll time Docker is up and answering, the Compose list is empty, `unhealthy.is_empty()` is true, and the instance is declared healthy while the image is still pulling.

An empty Compose container list now reports **not healthy** with reason `no container has been created yet`. "Docker answered but has nothing to show" is not the same as "the app is fine".

This composes correctly with the rule that containers without a `healthcheck` are not judged — the empty-list check separates *"Compose has not started"* from *"Compose started, and there is nothing to judge"*, which are the two cases that previously collapsed into "healthy".

I also split the decision rule out of `collect()` into a pure `judge()` so it is testable without a Docker daemon, and covered it:

- `no_compose_container_yet_is_not_healthy`
- `containers_without_a_healthcheck_are_not_judged`
- `every_declared_healthcheck_passing_is_healthy`
- `a_starting_container_is_not_healthy`
- `one_unhealthy_container_fails_the_instance`
- `docker_names_lose_their_leading_slash`

### 2. `enabled = false` did not restore the old behaviour — addressed in `5b466b09f`

Correct, and worse than the config comment suggested: in a fleet mid-upgrade the older instances are `Unsupported` (healthy), so `retain_healthy()` would keep only them and drop every new instance **permanently**.

Fixed by skipping the filter outright rather than merely leaving it unfed — a new `health_gating_enabled()` is consulted where the candidate's `healthy` flag is computed and in `instance_is_healthy`, so with polling off every instance is eligible regardless of what state it is sitting in.

Test `disabling_polling_keeps_both_old_and_new_instances_eligible` registers one legacy and one health-capable instance with polling disabled, asserts the capable one is still tracked as `Unknown` (the state is not faked), and that both are selected.

### 3. Also found: a WaveKV reload reset every health observation — `0847dfc49`

Same class, not in the review. `reload_instances_from_kv_store` rebuilt `InstanceInfo` and carried `connections` over from the existing record but not `health`, re-deriving it with `HealthState::initial()`. Since health is deliberately node-local and not in `InstanceData`, **every sync round reset the entire fleet to `Unknown`** — dropping every instance out of rotation until the next poll, repeatedly, which would have made health gating useless and churned routing.

`health` is now carried across a reload the same way `connections` is. The one case that still resets is `has_health_endpoint` changing, which means the image was replaced.

Test `a_kv_reload_does_not_reset_this_node_s_health_observations`.

### Note on the nerdctl limitation

The fix in (1) changes how the already-documented `nerdctl-compose` gap fails. Its containers are invisible to `Docker::connect_with_defaults()`, so the list is empty and every instance now reports unhealthy rather than healthy. Per app this is all-or-nothing — the runner comes from app-compose and is the same for every instance of an app — so fail-open catches it and routing is unchanged from today, but the dashboard will show `unhealthy` and the fail-open warning will fire. Still worth fixing properly with containerd support rather than leaving as-is.

`cargo test -p dstack-gateway --all-features`: 192 passed. `cargo test -p dstack-guest-agent`: 38 passed. `cargo fmt --check` and `clippy -D warnings` clean across gateway, guest-agent and dstack-util.

kvinwang (commented):
Self-review of this PR, posted as inline comments. Every finding below was verified against the code on `05d903c78e`; the ones marked with measurements were reproduced on this machine (nerdctl 2.3.5, docker 29.5.3) or by mutation-testing in a clean worktree.

**Blocking (P0)** — each of these can take a healthy instance out of rotation or let the CVM harm itself:

1. `app_health.rs` — a timed-out probe is not killed; one leaked root process per interval, unbounded.
2. `rpc_service.rs` — `Worker.Health` does unbounded work per unauthenticated request on the public 0.0.0.0:8090 listener.
3. `container_health.rs` — no timeout anywhere in `collect()`.
4. `container_health.rs` — a single container disappearing mid-collection destroys the entire report (both runners; measured on nerdctl 2.3.5).
5. `container_health.rs` — the judged set is "any compose project / the whole namespace", and container state survives reboots on the data disk ⇒ possible permanent unhealthy.
6. `proxy/health_check.rs` — unbounded response body from a guest-controlled peer, read every 5s.
7. `proxy/health_check.rs` — the round is applied as one batch with no generation guard, so a pre-reboot verdict can be written onto a post-reboot instance.

**Should fix (P1)** — `NERDCTL_NAMESPACE` override, missing `manifest_version` gate and `path` validation for `health_check`, unsanitised probe stderr, poll targets not filtered by handshake liveness, `has_health_endpoint` as `bool` instead of `Option<bool>`, and the vacuous random-path test.

**P2** — unsupervised poller task, no hysteresis, per-connection `warn_all_unhealthy` on the random path, debug-registration capability downgrade, zero tests for `health_check.rs`, and the false field-5 history claim in the proto comment (also in commit `54d7430b`'s subject).

Separately from the code: the PR description did not survive the rebase — "registration only runs at boot" is contradicted by this branch's own `cb7c8d8a96`; the "Yocto is still on nerdctl 2.2" limitation is stale (`origin/next`'s meta-virtualization is already `PV = v2.3.5`); the test counts are `195` / `53`, not `192` / `48`; and "Unknown is not healthy" is not *the* boot-window fix, because `retain_healthy` fails open for a single-instance app or a fleet booting together. `enabled = true` by default is also a routing-behaviour change for existing apps that already declare a compose `healthcheck` — a product call, not a bug, but it deserves an explicit rollout note.

dstack/guest-agent/src/app_health.rs
```diff
@@ -0,0 +1,245 @@
...
+            }
+        }
+        Ok(Err(err)) => {
+            return Verdict {
+                healthy: false,
+                reason: format!("failed to run {}: {err}", check.path),
+            }
+        }
+        Ok(Ok(output)) => output,
+    };
```
kvinwang: **P0 — a timed-out probe is never killed; one leaked root process per interval, unbounded.**

`tokio::process::Command` defaults to `kill_on_drop = false` ("a spawned process will, by default, continue to execute even after the `Child` handle has been dropped"). `tokio::time::timeout(...)` here only drops the `output()` future — the child keeps running. The loop then sleeps `interval_secs` and forks another one.

I verified this in a scratch test on this branch: two seconds after the timeout fires, `pgrep -f "sleep 31.5"` still finds the child.

No malice needed to hit it: an app probe that is `curl` against a wedged backend with no `--max-time` leaks a process every 10s → thousands a day → PID/memory exhaustion inside the CVM, which takes the app down with it. That is the exact opposite of what this feature is for.

Both the type doc (`dstack-types`: `timeout_secs` — "before it is **killed** and counted as a failure") and `docs/app-health-checks.md:84` ("A run that exceeds this is killed") promise behaviour the code does not implement.

Fix is one line: `command.kill_on_drop(true);` — plus a test that actually asserts the child is gone, since `a_hanging_probe_times_out_as_unhealthy` (:200) only checks the verdict string and is why this was invisible.

dstack/guest-agent/src/app_health.rs
```diff
@@ -0,0 +1,245 @@
...
+        // No code means a signal killed it.
+        None => "killed by signal".to_string(),
+    };
+    let stderr = String::from_utf8_lossy(&output.stderr);
+    let stderr = stderr.trim();
+    if stderr.is_empty() {
+        return code;
+    }
+    format!("{code}: {}", truncate(stderr))
+}
```
kvinwang: **P1 — app-controlled stderr crosses two trust boundaries unsanitised.**

Up to 512 bytes of the probe's stderr become `HealthResponse.unhealthy[].status`, which then:

1. is served to **anonymous callers on 0.0.0.0:8090** (see the `Worker.Health` comment on `rpc_service.rs`) — an app that prints a DSN or a token on failure publishes it to the internet; and
2. reaches the gateway's `info!("instance {id} is now {next} ...: {reason}")` with newlines and ANSI escapes intact.

`truncate()` bounds the length and cuts on a char boundary, but strips nothing. Contrast `http_routes.rs:267`, the agent's *other* app-output path, which does apply `strip_ansi_escapes`. Strip control characters here too before this crosses the boundary.

dstack/guest-agent/src/app_health.rs
```diff
@@ -0,0 +1,245 @@
...
+            latest: RwLock::new(Verdict {
+                healthy: false,
+                reason: "health check has not run yet".to_string(),
+            }),
+        };
+        let report = probe.report();
+        assert!(!report.healthy);
+        assert_eq!(report.unhealthy[0].name, PROBE_NAME);
+        assert_eq!(report.unhealthy[0].status, "health check has not run yet");
+    }
```
kvinwang: **Test quality — this does not test what its name claims.**

It hand-builds the `Verdict` instead of going through `spawn()`, so the invariant it is named for — that `spawn` seeds the cache as *unhealthy*, so an instance that has not probed yet is not routed to — is untested. Flip `spawn`'s initial `healthy: false` to `true` and this test still passes.

Driving it through `AppHealthProbe::spawn` and reading `report()` before the first tick would pin the real thing.

dstack/guest-agent/src/rpc_service.rs
```diff
@@ -624,6 +652,32 @@ impl WorkerRpc for ExternalRpcHandler {
...
+            Err(err) => {
+                warn!("failed to collect container health: {err:#}");
+                HealthResponse {
+                    healthy: false,
+                    unhealthy: vec![],
+                    error: format!("{err:#}"),
+                }
+            }
+        })
+    }
```
kvinwang, line 676: **P0 — unbounded work per unauthenticated request, on the public listener.**

This handler is on `ExternalRpcHandler`, mounted only by `run_external` (`server.rs:135-145`), which binds `address = "0.0.0.0"`, `port = 8090` (`guest-agent/dstack.toml:28-30`) — and `docs/security/cvm-boundaries.md:181` states that port "is publicly accessible".

Every call forks two root `nerdctl` processes (`container_health.rs:133,147`) or issues 1+N Docker API round trips (`container_health.rs:95-112`). No cache, no rate limit, no concurrency cap. An anonymous client can drive an unbounded number of concurrent `nerdctl ps`/`inspect` forks inside the CVM; enough of them starve the same Rocket server that `run_watchdog` probes (`server.rs:340-348`, `WatchdogSec=30s`, `Restart=always`), so the flood escalates into guest-agent restarts.

The irony is that `app_health.rs:12-16` explicitly reasons that a fleet of gateway pollers "must not multiply into that many executions of the app's probe" — and then the container path is given exactly that property, *plus* an anonymous public trigger.

Make the container verdict run on its own timer with a short cache and have this RPC read the cached conclusion, exactly like the `app_health` branch two lines above.

dstack/guest-agent/src/container_health.rs
```diff
@@ -0,0 +1,601 @@
...
+
+/// The label `docker compose` stamps on every container it starts.
+const COMPOSE_PROJECT_LABEL: &str = "com.docker.compose.project";
+
+/// The containerd namespace `app-compose.sh` runs nerdctl in.
+///
+/// Kept in sync with `NERDCTL_NAMESPACE` there. The namespace is dedicated to
+/// the app, which is why the nerdctl path needs no label filter: everything in
+/// it belongs to the app by construction.
+const NERDCTL_NAMESPACE: &str = "dstack";
```
kvinwang: **P1 — the namespace is hardcoded here while the runner honours an env override.**

`os/common/rootfs/app-compose.sh:13` is `NERDCTL_NAMESPACE="${NERDCTL_NAMESPACE:-dstack}"`, and `app-compose.service` loads `EnvironmentFile=-/dstack/.host-shared/.decrypted-env`. So a deployer who sets `NERDCTL_NAMESPACE` in the app env starts containers in namespace X while the agent lists namespace `dstack`, gets an empty list, and reports `healthy: false, "no container has been created yet"` **forever** — the instance never enters rotation, with a message that says the opposite of what happened.

Either read the same env var here, or drop the override from `app-compose.sh` so the constant is actually a constant.

dstack/guest-agent/src/container_health.rs
```diff
@@ -0,0 +1,601 @@
...
+/// that exited cleanly.
+pub(crate) async fn collect(runner: &str) -> Result<HealthReport> {
+    match runner {
+        "docker-compose" => collect_docker().await,
+        "nerdctl-compose" => collect_nerdctl().await,
+        // The `bash` runner runs a script, not containers. There is nothing to
+        // inspect and never will be, so it must not sit at "not started yet".
+        _ => Ok(HealthReport::not_judged()),
+    }
+}
```
kvinwang: **P0 — no timeout anywhere in `collect()`.**

Neither `collect_docker` (:91-130) nor `nerdctl()` (:156-169) wraps its I/O in `tokio::time::timeout`. The gateway's 2s client-side timeout (`gateway/src/proxy/health_check.rs:109`) cancels only the *gateway's* future — the guest-side `nerdctl` child is neither killed (no `kill_on_drop` here either) nor bounded.

With containerd wedged and a 5s poll from every gateway node, in-flight handlers and orphaned `nerdctl` children accumulate monotonically inside the CVM.

Also worth noting: the `_ => not_judged()` fail-open arm on line 87 means any future runner name silently becomes "always healthy", and nothing covers this dispatch at all.

dstack/guest-agent/src/container_health.rs
```diff
@@ -0,0 +1,601 @@
...
+    let mut filters = HashMap::new();
+    filters.insert("label", vec![COMPOSE_PROJECT_LABEL]);
+    let containers = docker
+        .list_containers(Some(ListContainersOptions {
+            all: true,
+            filters,
+            ..Default::default()
+        }))
+        .await
+        .context("failed to list containers")?;
```
kvinwang, line 181: **P0 — the verdict surface is far wider than "this app", and container state survives reboots.**

Two halves:

- Docker: the filter is `label=com.docker.compose.project` **with no value** — i.e. "has a compose project label", not "belongs to *this* project".
- nerdctl (:133): no filter at all, justified at :34-36 as "everything in the namespace belongs to the app by construction".

That construction argument does not hold, because `/var/lib/docker`, `/var/lib/containerd` and `/var/lib/nerdctl` are rbind-mounted from the **persistent data disk** (`os/common/rootfs/dstack-prepare.sh:285-299`). Container records survive reboots and app upgrades. The only pre-start cleanup, `dstack-util remove-orphans --no-dockerd` (`dstack-prepare.sh:339`), is Docker-only *and* skips anything whose `com.docker.compose.project` differs from the directory-derived name (`docker_compose.rs:101-104`); `nerdctl compose up --remove-orphans` likewise only touches its own project.

Concrete failure: an app whose compose file carries a top-level `name:` (or that ran its own `nerdctl run` from `pre_launch_script`) upgrades to a version without it. The old exited container that declared a `healthcheck` stays on the data disk, `judge` sees `exited (last health check: healthy)`, and the instance is **permanently** out of rotation with no in-app remedy short of wiping the disk.

This is amplified by the (correct) "not running ⇒ unhealthy" rule added in `86a69a93fe`. The rule is right; the judged set needs narrowing to the current compose project.

dstack/guest-agent/src/container_health.rs
```diff
@@ -0,0 +1,601 @@
...
+
+    let mut observed = Vec::with_capacity(containers.len());
+    for summary in containers {
+        let Some(id) = summary.id.as_deref() else {
+            continue;
+        };
+        let details = docker
+            .inspect_container(id, None)
+            .await
+            .with_context(|| format!("failed to inspect container {id}"))?;
```
kvinwang: **P0 — one container disappearing mid-collection destroys the whole report.**

`inspect_container(id)?` aborts `collect_docker` on the first 404. `docker compose up` recreates containers on every redeploy/restart, and `app-compose.sh:105-106` runs `docker image prune -af` / `volume prune -f` right after start, so the list→inspect window is routinely raced.

Result: `HealthResponse { healthy: false, error: "failed to inspect container <id>: ..." }` and the instance leaves rotation for that round for no reason.

Skip containers that vanished between `list` and `inspect` instead of failing the batch.

dstack/guest-agent/src/container_health.rs
```diff
@@ -0,0 +1,601 @@
...
+
+    let mut args = vec!["inspect"];
+    args.extend_from_slice(&ids);
+    let raw = nerdctl(&args)
+        .await
+        .context("failed to inspect nerdctl containers")?;
+    let containers: Vec<NerdctlContainer> =
+        serde_json::from_str(&raw).context("failed to parse nerdctl inspect output")?;
+
+    Ok(judge(containers.into_iter().map(observe_nerdctl).collect()))
```
kvinwang: **P0 — same failure as the Docker path, confirmed on real nerdctl.**

`nerdctl inspect` exits non-zero if **any** id in the batch is gone, and `nerdctl()` (:164-167) turns a non-zero exit into a hard error for the whole batch. Measured on nerdctl 2.3.5 on this machine:

```
$ nerdctl --namespace hctest inspect 6c612929c546 deadbeefdead
exit=1
level=fatal msg="1 errors: [no such object deadbeefdead]"
```

One id recycled between `ps -a -q` and `inspect` — compose recreating a container, a prune — and `collect()` bails, the report is `healthy: false`, and the instance drops out of rotation for the round.

Judge the containers that *did* come back rather than failing the whole batch.

dstack/guest-agent/src/container_health.rs
```diff
@@ -0,0 +1,601 @@
...
+}
+
+async fn nerdctl(args: &[&str]) -> Result<String> {
+    let output = Command::new("nerdctl")
+        .arg("--namespace")
+        .arg(NERDCTL_NAMESPACE)
+        .args(args)
+        .output()
```
kvinwang: **Speculative, cheap to settle: `nerdctl inspect` is invoked without `--snapshotter`.**

`app-compose.sh:65,78,109` always passes `--snapshotter "$snapshotter"`, and the value is available to the agent as `config.app_compose.snapshotter`. If nerdctl's dockercompat inspect resolves snapshot data through the global flag rather than the container record, every stargz-snapshotter app gets a hard collector error and permanent unhealthy.

The captured fixture at :471 is overlayfs only, so the test suite would not catch it. Worth a single run on stargz hardware before merge.

dstack/dstack-types/src/lib.rs
```diff
@@ -225,6 +225,14 @@ pub struct AppCompose {
...
     )]
     pub init_script: Vec<String>,
+    /// An app-declared health probe, run periodically by the guest agent.
+    ///
+    /// When present it replaces whatever the runner would otherwise report --
+    /// including the container runners, whose per-container `healthcheck`
+    /// declarations it overrides. For a runner that has no containers to
+    /// inspect (`bash`) it is the only way to report health at all.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub health_check: Option<HealthCheck>,
```
kvinwang: **P1 — `health_check` breaks the project's own fail-closed convention, and is never validated.**

`verify_manifest_feature_requirements` (`dstack-util/src/system_setup.rs:982-1002`) gates `requirements`, `nerdctl-compose` and multi-`init_script` behind manifest v3, each with the comment *"so older guests fail closed"*. `health_check` has no such gate, and `AppCompose` does not `deny_unknown_fields` — so an app-compose that declares a probe deploys onto an older guest image and is **silently ignored**, the operator's only signal being the absence of a log line.

Separately there is no validation of the field at all: `path: ""`, or a path that does not exist, deserializes fine and yields a permanently unhealthy instance discovered only from gateway logs.

docs/app-health-checks.md
```diff
@@ -0,0 +1,133 @@
...
+  "manifest_version": 1,
+  "name": "my-app",
+  "runner": "bash",
+  "bash_script": "/opt/my-app/run.sh &",
+  "health_check": {
+    "path": "/opt/my-app/healthz.sh",
+    "args": ["--strict"],
+    "interval_secs": 10,
+    "timeout_secs": 5
+  }
```
kvinwang: **Doc defect with an availability consequence: the probe runs in the guest rootfs, not in the app's container.**

`app_health.rs:105` spawns `check.path` as a child of the guest agent — outside every container, as root, with access to `/dstack`. This page never says so, and actively encourages compose apps to declare one (":93-95": "overrides container healthchecks, so a Compose app that wants one whole-application answer can give one").

The example path `/opt/my-app/healthz.sh` is exactly the shape that lives *inside the image* for a compose app and does not exist in the guest rootfs. Following the example verbatim yields `failed to run ...: No such file or directory` on every tick and an instance that never serves.

The page needs to state the execution context explicitly: guest rootfs, guest binaries only (`curl` and `busybox-static` are present per `os/mkosi/mkosi.conf:56,75`).

dstack/gateway/rpc/proto/gateway_rpc.proto
```diff
@@ -130,6 +142,10 @@ message ProxyAccelStatus {
 
 // HostInfo is the information of a host.
 message HostInfo {
+  // Field 5 was used and dropped before this repo's visible history, without a
+  // marker. Reserved now so the next person to extend this message cannot
+  // silently reuse it and break every older client that still sends it.
```
kvinwang: **This history claim is false, and it is also in the commit subject of `54d7430b`.**

Field 5 was `uint32 port`. It is visible at `62e876ce73` and `00def34b5c`, and was removed by `b325c0f8bf` ("feat(gateway): support multi-port TCP listening via port ranges", 2026-02-02), whose own message says *"Also removes the single-port field from HostInfo proto"*. That is squarely inside this repo's visible history.

Reserving the number is still the right call — but state the real reason (dropped by `b325c0f8bf` when the single-port field went away) rather than an unverifiable one, and fix the commit subject to match.

dstack/gateway/src/proxy/health_check.rs
```diff
@@ -0,0 +1,149 @@
...
+    tokio::spawn(async move {
+        let mut ticker = tokio::time::interval(config.interval);
+        // A round that overruns the interval must not leave catch-up rounds
+        // queued behind it; skip to the next tick instead.
+        ticker.set_missed_tick_behavior(MissedTickBehavior::Delay);
+        loop {
+            ticker.tick().await;
+            poll_round(&state, &config).await;
+        }
+    });
```
kvinwang, line 128: **P2 — bare unsupervised `tokio::spawn`: if this task ever dies, polling stops silently and forever.**

Nothing logs it, and nothing on the dashboard or the admin API distinguishes "polled and healthy" from "frozen at the last verdict". Note `state.lock()` is `or_panic` (`main_service.rs:250`), so a poisoned `ProxyState` mutex kills this task.

The consequence is not benign: every *subsequently* registered instance stays `Unknown` forever, and `Unknown` is only rescued by fail-open when the app has **no** healthy peer. An app with one legacy (`Unsupported` ⇒ healthy) or one stuck-`Healthy` instance then pins 100% of its traffic onto that one instance indefinitely.

Supervise the task (restart + `error!`), or at minimum expose a "last completed round" timestamp so the freeze is observable.

dstack/gateway/src/proxy/health_check.rs
```diff
@@ -0,0 +1,149 @@
...
+        let guard = state.lock();
+        guard
+            .state
+            .instances
+            .values()
+            // Instances that never declared the capability cannot answer, so
+            // polling them would only produce failures to misread.
+            .filter(|instance| instance.has_health_endpoint)
+            .map(|instance| (instance.id.clone(), instance.ip))
+            .collect::<Vec<_>>()
```
kvinwang: **P1 — poll targets are not filtered by liveness; dead instances eat the whole round budget.**

The only filter is `has_health_endpoint`. But `recycle.timeout = "10h"` (`gateway.toml:175`) and `handshake_stale = "30m"` (`:154`), so a powered-off CVM sits in `state.instances` for up to ten hours, is **already excluded from routing** by the handshake filter, and is still polled every round — burning a full `timeout = 2s` slot each time.

With the shipped defaults (`concurrency = 16`, `timeout = 2s`, `interval = 5s`), ~40 timing-out polls fit in one 5s round:

| dead instances | round duration |
|---|---|
| 200 | ~25s |
| 1000 | ~125s |

`MissedTickBehavior::Delay` then stretches the effective interval to match, so a handful of dead CVMs degrades health-detection latency **for every live instance in the fleet** — and, via the batching issue below, the staleness of every applied verdict. A tenant can trigger this deliberately by registering CVMs that declare `has_health_endpoint` and then accept-and-hang.

Filtering targets by `handshake_stale` before polling removes almost all of it.

dstack/gateway/src/proxy/health_check.rs
```diff
@@ -0,0 +1,149 @@
...
+            (instance_id, poll_instance(ip, agent_port, timeout).await)
+        })
+        .buffer_unordered(config.concurrency.max(1))
+        .collect::<Vec<_>>()
+        .await;
+
+    let mut guard = state.lock();
+    for (instance_id, observation) in observations {
+        guard.record_instance_health(&instance_id, observation);
+    }
```
kvinwang: **P0 — the whole round is buffered and applied in one batch, with no generation guard. This defeats the feature's headline guarantee.**

`.collect::<Vec<_>>().await` buffers every observation until the *last* one lands; `record_instance_health` (`main_service.rs:1520-1523`) then only checks that the instance still exists.

```
t0  round starts; X is snapshotted and polled -> healthy
t1  X reboots, re-registers with a fresh WG key -> health reset to Unknown
    (exactly the reset `a_reboot_resets_health_to_unknown` pins)
t2  round finishes -> the t0 observation is applied -> X = Healthy
```

X now serves traffic on the strength of a poll of its **previous boot**, containers not yet up — precisely what `a_newly_registered_instance_waits_for_its_first_health_poll` says must not happen. The window is not `timeout`; it is *the rest of the round*, which grows with fleet size (see the liveness-filter comment above).

Same hole for IP reuse: the snapshot carries `(id, ip)` (:78), so if the instance is rebuilt with a different IP (`new_client_by_id`'s invalid-IP rebuild, or a reload writing a changed `data.ip`), the verdict polled from the old address is recorded against the new record.

Both are fixed by the same change: apply each observation as it completes, and carry a generation token — the WireGuard public key would do — into the snapshot and compare it in `record_instance_health`.

dstack/gateway/src/proxy/health_check.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Ask one agent whether its app is serving.
+///
+/// Any failure to get an answer counts as unhealthy. That is safe to be strict
+/// about because the caller fails open when it would empty an app's candidate
+/// set, and because instances whose agent cannot answer at all were filtered
+/// out before we got here.
+async fn poll_instance(ip: Ipv4Addr, agent_port: u16, timeout: Duration) -> Observation {
+    let url = format!("http://{ip}:{agent_port}/prpc");
+    let client = WorkerClient::new(PrpcClient::new(url));
+    let response = match tokio::time::timeout(timeout, client.health()).await {
```
kvinwang: **P0 — unbounded response body from a guest-controlled peer, now read on a 5s timer.**

`PrpcClient::request` → `http-client/src/lib.rs:74-87` does `response.text().await?` for an `http://` base with **no size cap** (and a second UTF-8 allocation on top).

`has_health_endpoint` is *self-declared by the CVM* (`gateway_rpc.proto` field 3), so a compromised guest opts **itself** into being polled every interval and then answers `Worker.Health` with an arbitrarily large body. The only bound is the 2s timeout: at tunnel line rate that is hundreds of MB per poll, × `concurrency = 16`, every 5s.

This codebase already knows this failure mode and fixed it for the peer-facing client: `kv/https_client.rs:31-42` (`read_body_bounded`, `Limited::new(body, MAX_COMPRESSED_SYNC_BYTES)`) with the comment *"a peer could stream an unbounded response … the memory is already gone by then"*, plus the test `an_oversized_response_body_is_refused`. `port_policy::fetch_port_policy` shares this transport but fires lazily, once per instance; this PR turns it into a repeating, guest-triggerable allocation against the whole fleet.

Apply the same bound here.

dstack/gateway/src/proxy/health_check.rs
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+        Ok(Err(err)) => {
+            return Observation {
+                state: HealthState::Unhealthy,
+                reason: format!("health poll failed: {err:#}"),
+            }
+        }
+        Ok(Ok(response)) => response,
+    };
```
kvinwang, line 327: **P2 — no hysteresis: a single 2s timeout ejects the instance *and* drops the app's cached selection.**

One timeout or one transport error is immediately `Unhealthy`; `record_instance_health` then does `top_n.remove(&app_id)` (`main_service.rs:1545`). Against an agent that is momentarily busy — image pull, log flush, CPU-starved CVM — the instance leaves rotation for at least one interval and the whole app's selection is recomputed. Both edges of a flap cost a cache invalidation.

Nothing requires N consecutive failures before demoting. `config.rs:493-495` explicitly reasons about Docker's own `retries` for the *guest* side, but the transport side gets none.

Related, in the same file: a fresh `reqwest::Client` (pool, resolver, TLS config) is built per poll inside `http-client/src/lib.rs:76`. Pre-existing in the transport, but this PR is the first caller to run it on a fixed timer against the entire fleet.

dstack/gateway/src/proxy/health_check.rs
```diff
@@ -0,0 +1,149 @@
+// SPDX-FileCopyrightText: © 2024-2025 Phala Network <dstack@phala.network>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//! Application-level health polling of registered CVMs.
```
kvinwang, line 5: **All 149 lines of this file are untested.**

Target selection, the `has_health_endpoint` filter, the batching, `poll_instance`'s error→`Unhealthy` classification, and `describe_unhealthy` have no coverage. The `observe` helper in `tests.rs:498-506` calls `record_instance_health` directly and never exercises the poller.

That is also where four of the findings on this PR live (unbounded body, batch-without-generation, no liveness filter, no hysteresis) — the untested half is not a coincidence.

dstack/gateway/src/main_service.rs
```diff
@@ -1179,6 +1194,40 @@ fn reload_instances_from_kv_store(proxy: &Proxy, store: &KvStore) -> Result<()>
...
+}
+
+/// Bounded by the `cache_top_n` recomputation interval, so an app stuck in this
+/// state costs a couple of lines a minute rather than one per connection.
+fn warn_all_unhealthy(app_id: &str, candidates: usize) {
+    warn!(
+        "app {app_id}: no instance reports healthy; routing to all {candidates} \
+         reachable instance(s) anyway rather than refusing the app"
+    );
+}
```
kvinwang: **P2 — this doc comment is false on the random path, which is where the flood happens.**

"Bounded by the `cache_top_n` recomputation interval … rather than one per connection" is true on the top-n path, where the fail-open result is cached (`:1775-1777`). It is **not** true on the random path: `random_select_a_host` (`:1787`) has no cache and runs per connection — both when `connect_top_n = 0` and via the handshake-error fallback at `:1734`/`:1747`.

So an app whose whole fleet is unhealthy — the fail-open case, which is the *sustained* one — emits one `warn!` per inbound connection. That is exactly the log flood the comment claims to prevent.

dstack/gateway/src/models.rs
```diff
@@ -49,10 +49,99 @@ pub struct InstanceInfo {
     /// of its own.
     #[serde(default)]
     pub admin_ready: Option<bool>,
+    /// Whether the guest agent implements `Worker.Health`, as declared at
+    /// registration. Persisted, so a gateway restart does not start polling a
+    /// legacy image and conclude from the failures that it is unhealthy.
+    #[serde(default)]
+    pub has_health_endpoint: bool,
```
kvinwang: **P1 — plain `bool` here abandons the presence-tracking that the field directly above it exists to have.**

`admin_ready` is `Option<bool>` precisely so "an older node dropped this" is distinguishable from "set to false" — `gateway_rpc.proto:161-164` spells the reasoning out. `has_health_endpoint` is `#[serde(default)] bool` in the same WaveKV record.

In a mixed-version cluster, a pre-PR node that rewrites an `inst/` record (it does, on every re-registration it handles — `main_service.rs:1354`) silently zeroes the flag. On a new node the reload then sees `existing.has_health_endpoint != data.has_health_endpoint` (`:1143-1145`), resets health to `Unsupported`, and — via the new `health` clause in `routing_inputs_differ` (`models.rs:189`) — drops the app's cached selection. A CVM re-registering round-robin across old and new nodes flaps `Unsupported` ↔ `Unknown` and invalidates `top_n` on every sync round.

`Option<bool>` with "absent ⇒ leave the current capability alone" avoids it.

dstack/gateway/src/debug_service.rs
```diff
@@ -36,7 +36,12 @@ impl DebugRpc for DebugRpcHandler {
             &request.instance_id,
             &request.client_public_key,
             "",
-            None,
+            // Reports nothing about itself: no port policy, and no health
+            // endpoint, so a debug-registered instance is never polled and
+            // always counts as healthy. That keeps this path usable for
+            // testing routing without standing up a guest agent, but it also
+            // means health gating cannot be exercised through it.
+            Default::default(),
```
kvinwang: **P2 — this can permanently downgrade an instance that a *real* CVM registered.**

`Default::default()` ⇒ `has_health_endpoint: false`. If an instance id already registered by a real CVM (with the capability) is re-registered through this path, `main_service.rs:1310-1313` sees the capability change, resets health to `Unsupported`, and `health_check.rs:74` then excludes it from polling **forever**.

The comment (and the commit message) documents "debug registration bypasses health gating" for *new* instances; neither notes that it can silently downgrade an existing one. Worth either preserving the existing capability on re-registration through this path, or saying so here.

dstack/gateway/src/main_service/tests.rs
```diff
@@ -751,6 +856,402 @@ async fn gating_an_unregistered_instance_is_an_error() {
...
+    );
+}
+
+/// The two selection paths express the "polling off means no gating" rule
+/// through different mechanisms -- the top-n path folds it into the candidate
+/// flag, the random path checks it inside `instance_is_healthy` -- so a
+/// refactor could easily fix one and miss the other. This pins the combination
+/// the other disabled-polling test does not reach.
+#[tokio::test]
+async fn disabling_polling_also_keeps_the_random_path_open() {
```
kvinwang: **This test is vacuous with respect to the mutation its own doc comment names.**

The comment states the purpose: *"the random path checks it inside `instance_is_healthy` … a refactor could easily fix one and miss the other."* I deleted exactly that check:

```rust
// instance_is_healthy, main_service.rs:1705-1707
- if !self.health_gating_enabled() {
-     return true;
- }
```

Result: **195 passed, 0 failed.** The test does not catch the mutation it was written for.

Cause: `register_instances(&mut proxy, "off-rand-app", 2, true)` makes *both* instances `Unknown`, so `any_healthy` is false and the **fail-open** branch (`main_service.rs:1819-1825`) keeps them — masking the disable switch entirely.

The top-n twin `disabling_polling_keeps_both_old_and_new_instances_eligible` uses a mixed `[false, true]` fleet and *does* catch the analogous mutation. Give this one the same mixed fleet: one `Unsupported` peer is what makes `any_healthy` true and actually exercises the branch.

There is a real (currently latent) asymmetry underneath: on the random path, "polling disabled" is load-bearing only for mixed fleets, and nothing pins it.

kvinwang:
Pushed `a1c2cad536`. Every inline finding above is addressed; the feature is now opt-in and `Worker.Health` costs a lock and a clone. The description is rewritten to match.

The design changed rather than being patched in place:

- **Opt-in under `requirements`** — `requirements.health_check = {enabled, health_file}`. `Requirements` is `deny_unknown_fields` and already gated at `manifest_version >= 3`, so the missing manifest gate is solved structurally instead of by one more `bail!`. An app that does not opt in is never polled.
- **The program probe is gone.** `health_file` is two lines the app writes; reading a file cannot hang and cannot leak a process, so ①'s `kill_on_drop` and most of the timeout machinery stop being needed at all.
- **The verdict is computed on the agent's own 5s timer** and the RPC reads the cache — ② and ③.
- `has_health_endpoint` → `health_check`, `HealthState::Unsupported` → `Ungated`.

Two things I measured rather than assumed, both of which changed the code:

- **nerdctl sets only `com.docker.compose.project`/`.service`/`.config-hash`.** The docker-only `project.working_dir` label I first reached for would have matched *nothing* under the runner the guest image actually uses — an empty list, which `judge` reads as "nothing started yet", i.e. permanently unhealthy.
- **A batch `nerdctl inspect` with one stale id exits 1 with *empty* stdout**, wherever the bad id sits, while docker still returns the entries it found. So ④ needs a per-id fallback, not "parse stdout anyway".

I then ran an adversarial pass over the fixes, which found ten more — including a regression the first round introduced. Worth calling out three:

- **A transport-wide 1 MiB body cap broke `dstack vmm logs`.** Same helper, and `--lines` is unbounded. The bound is now the caller's choice (`PrpcClient::with_max_response_bytes`), set on the two gateway→guest-agent clients only.
- **⑦'s generation guard had a slower twin.** The reboot reset is applied by whichever node took the re-registration; every *other* node learns about the new boot through KV reload, which carried its own pre-reboot verdict across on the capability alone. The guard closed a 2s window while a whole-interval one stayed open on every peer.
- **The snapshot's default path does not exist in the shipped gateway CVM.** `entrypoint.sh` emitted no `[core.proxy.health_check]` section at all, so a new module would have been silently inert exactly where it was needed.

Also: the SDKs' `Requirements` is a closed shape, so Go dropped `health_check` on marshal and Python raised — meaning the SDK would hand back a compose hash for an app-compose that is not the one being deployed, and that hash is what gets whitelisted on chain. Fixed in Go, Python and JS with tests. (`requirements.gpu_policy` has the same hole; pre-existing, left alone.)

Every new test was mutation-verified, and `disabling_polling_also_keeps_the_random_path_open` — vacuous before — now fails on the mutation it was written to catch. `223 / 67 / 95 / 39 / 9` passing, fmt and clippy clean, each of the four commits compiles on its own.

One thing I did **not** change, since it is not a hole: `NERDCTL_NAMESPACE` stays overridable by the app. Everything the app does is measured, so an app that moves its own containers out of view is making a measured choice about itself. The constant's doc says so now.

kvinwang:
Second adversarial pass — three reviewers in isolated worktrees, one lens each (the previous round's fixes / end-to-end wiring / hostile-guest and resources). 21 findings, all addressed in `205f0dd520`. Two of them are attack surface the *previous* round introduced, which is the argument for doing this twice.

**`health_file` was an arbitrary-file read whose contents reached an anonymous caller.** `tokio::fs::File::open` follows symlinks and the agent runs as root. The path is measured into the compose hash; what sits *at* that path at runtime is not — and the docs tell the app to bind-mount the directory into a container, so the container can swap it for a symlink afterwards. The parser then quoted what it found back: a line that failed to parse as a timestamp returned as `"...is not a unix time in seconds: <that line>"`, 64 bytes at a time, on the unauthenticated `0.0.0.0:8090` listener. Now `O_NOFOLLOW`, regular-file only, and the contents are never quoted — a verdict names the rule that failed.

**A FIFO at that path killed the agent's blocking pool, silently.** `tokio::fs` opens on `spawn_blocking`, and a blocking task cannot be cancelled, so `open(2)` on a reader-less FIFO parks that thread forever while the timeout merely drops the future. One per refresh → pool dead in ~2h → no filesystem work anywhere in the root agent. The watchdog would not catch it; it probes over HTTP and never touches the pool. `O_NONBLOCK` fixes it, and the test for it *hangs* if you remove the flag — which is how it was confirmed.

**`COMPOSE_PROJECT_NAME` defeated the container path entirely.** Measured here on docker compose 5.1.4 and nerdctl 2.3.5: **both honour it, and it outranks the compose file's top-level `name:`.** The app supplies it via `.decrypted-env`, so deriving the project from the file was wrong exactly when an app sets one — empty container list → "no container has been created yet" → permanently unhealthy, with a message saying the opposite of what happened. `NERDCTL_NAMESPACE` had the same shape.

The fix inverts it: `app-compose.sh` records the namespace and project it *actually used* (resolved by `docker compose config`, which applies the real precedence) and the agent reads that. Its absence is meaningful too — written before anything starts, so "not there yet" is genuinely the boot window. That also **deletes** the YAML parsing this branch added, along with its dependency.

**The poll round had no budget.** Measured against a listener that accepts and never answers: `ceil(n/16) × 2s`, so 256 mute instances make a 33s round — and `MissedTickBehavior::Delay` then makes *that* the interval. One tenant's mute CVMs become every other tenant's detection latency, permanently, because being marked unhealthy does not remove an instance from the target list. The round now gets one interval, rotates where it starts so nothing starves, and logs what it did not reach.

Smaller, in one line each: one flapping instance rewrote the whole fleet's snapshot every round (~39 KB/s at 2000 instances) → debounced; `select_targets` held the routing mutex across a call that clones every peer's key and can `fork`+`exec` `wg show` → moved outside; `warn_all_unhealthy`'s rate limit was a process-global mutex on the per-connection path, so one tenant's fail-open serialized against another's connections → moved into state the caller already holds; `sanitize` mixed char counts and byte counts three ways *and* `char::is_control` misses U+2028 and the bidi overrides → one shared, tested implementation; `HealthStore::restore` could leave an instance out of rotation *and* out of the poll set with nothing able to lift it; the VMM's vsock client to a guest agent was unbounded; deploy-time validation missed the likelier mistake (`enabled` with no `healthcheck:` anywhere in the compose file); Go could not express `health_file: ""` and so hashed differently from the other three SDKs.

Four things a mutation showed were untested now are — including the `requirements.health_check.enabled` → `RegisterCvmRequest` hop, which is the *only* wire between the manifest and the gateway: making it inert left all 95 `dstack-util` tests green. Two others asserted a constant against itself.

Three documentation statements were wrong, not merely thin: `requirements` only fails an older guest closed if you write `"manifest_version": "3"` **as a string** (`AppCompose` is not `deny_unknown_fields`); a single-file bind mount and write-then-rename cancel each other out, because the rename never touches the inode the agent holds; and the changelog claimed a transport-wide 1 MiB cap that was deliberately not what shipped.

`227 / 71 / 100 / 46 / 9` passing, clippy and prek clean, each commit compiles standalone.

PR #1080, opened 2026-08-20
feat(guest-agent): read container health from nerdctl as well as docker

kvinwang:
> Stacked on #1079. Closes the `nerdctl-compose` gap that PR listed under "Known limitations".

## Problem

#1079 only covers the `docker-compose` runner. `app-compose.sh` also supports
`nerdctl-compose`, whose containers live in a containerd namespace rather than
behind a Docker-compatible socket, so `Docker::connect_with_defaults()` sees
nothing. With the empty-list rule from that PR, every nerdctl instance now
reports unhealthy forever.

That is the safe direction — per app the runner is uniform, so fail-open catches
it and routing matches today's behaviour — but the feature does nothing there,
and the dashboard shows a fleet of red.

There are really two separate gaps, and they need different fixes:

**A. Visibility.** The guest agent cannot see nerdctl containers at all. This is
independent of the nerdctl version.

**B. Compose `healthcheck:` being honoured.** `nerdctl compose` did not parse
the field until recently. From the tagged sources:

| nerdctl | `pkg/composer/serviceparser` handles `HealthCheck` |
|---|---|
| 2.2.1 (was pinned here) | no — falls through to the "Ignoring: service ..." warning |
| 2.3.0 | no |
| **2.3.1** | **yes** — translated into `--health-cmd` / `--health-interval` / `--no-healthcheck` |
| 2.3.5 (latest) | yes |

(`nerdctl run --health-cmd` itself landed in 2.1; what was missing was the
Compose-layer translation. Issue containerd/nerdctl#2386 is still open but stale
— the code is there.)

## Fix

**A — read health from nerdctl.** `container_health::collect` now dispatches on
`AppCompose.runner`:

- `docker-compose` — unchanged, over the Docker socket.
- `nerdctl-compose` — `nerdctl --namespace dstack ps -a -q`, then
  `nerdctl inspect` on the result. nerdctl generates its inspect output from its
  own `dockercompat` types, so `State.Health.Status` carries the *same four
  strings* Docker uses (`none` / `starting` / `healthy` / `unhealthy`). The
  decision rule is genuinely shared rather than reimplemented — `judge()` now
  takes the status as a string and both collectors feed it.
- `bash` — runs a script, not containers. Not judged, so it does not sit at
  "not started yet" forever.

No label filter is needed on the nerdctl path: `app-compose.sh` runs it under a
dedicated `dstack` namespace, so everything in it belongs to the app by
construction.

`judge()` also now fails closed on an unrecognised status. Previously the match
was exhaustive over bollard's enum; against a string from an external CLI it has
to assume a state it does not know about is not "ready".

Uses `tokio::process` rather than `cmd_lib` so the poll does not block the
runtime — the feature is enabled the same way `kms/Cargo.toml` already does it.

**B — nerdctl 2.3.5 on the mkosi backend.** `versions.env` and the acceptance
pin. The checksum is the upstream one:

```
$ curl -sL .../v2.3.5/SHA256SUMS | grep linux-amd64.tar.gz
de3206aeb7cbd5f20f5fb1f55c1e3bf2db1be567812a8a3f5e65eba2488347ee  nerdctl-2.3.5-linux-amd64.tar.gz
```

## Not done here: the Yocto backend

**Yocto is the default backend and it is still on nerdctl release/2.2**, so (B)
does not reach production images with this PR. (A) does, and on its own restores
the boot-window gate for nerdctl — an empty container list is still an empty
container list whether or not healthchecks are configured.

Bumping Yocto is not a version-string change:

- The recipe lives in the `Dstack-TEE/meta-virtualization` submodule
  (`dstack-wrynose`), not in this repo, so it needs its own PR plus a submodule
  pointer bump here.
- `nerdctl_git.bb` pins `SRCREV` and `branch=release/2.2`, and the build is fed
  by ~1670 lines of generated Go module manifests
  (`go-mod-hybrid-gomod.inc`, `go-mod-git.inc`, …). Per
  `QUICKSTART-oe-go-mod-vcs.md` these are regenerated from a BitBake *discovery
  build* — populate `GOMODCACHE`, then run `extract-discovered-modules.py`.
  That needs a Yocto build host; hand-editing them would break reproducibility
  of the production image.
- It changes the guest OS measurements, so it carries a KMS image-hash rollout.

Happy to do it as a follow-up given a build host.

## Verification

`cargo test -p dstack-guest-agent`: 40 passed (8 new). `cargo test -p dstack-gateway --all-features`: 192 passed.
`cargo fmt --check` and `clippy -D warnings` clean.

The shared rule is covered directly, including the two cases the runner split
introduces:

- `nerdctl_inspect_output_parses` — pins the field names and the
  Health-absent-when-no-healthcheck shape against a real `nerdctl inspect`
  payload
- `an_unrecognised_state_holds_traffic_back` — fail-closed on unknown status
- plus the existing `no_container_yet_is_not_healthy`,
  `containers_without_a_healthcheck_are_not_judged`,
  `a_starting_container_is_not_healthy`, `one_unhealthy_container_fails_the_instance`,
  `every_declared_healthcheck_passing_is_healthy`, `docker_names_lose_their_leading_slash`

**Verified on hardware** (tdx-lab: containerd 2.2.1, systemd 255, nerdctl 2.3.5),
which closes both of the things I could not confirm from source:

*1. `nerdctl compose` really does translate the healthcheck.* From the run log:

```
level=info msg="Running [nerdctl --namespace=nchp run ... --name=nchp-withcheck-1
  --health-cmd=true --health-interval=2s --health-timeout=2s
  --health-retries=2 --health-start-period=1s docker.io/library/busybox:latest ...]"
```

Worth knowing: it warns `healthcheck: CMD exec form is not supported, converting
to CMD-SHELL`. Compose files using the `CMD` exec form get silently reshaped
into `CMD-SHELL` on nerdctl. Not a problem for this PR — the verdict is the same
either way — but it is a behavioural difference from Docker.

*2. The checks actually run, via systemd timers, and inspect reports the result.*
A three-service project (passing check / no check / failing check):

```
Name=nchp-withcheck-1   State.Status=running  Health={'Status': 'healthy',   'FailingStreak': 0}
Name=nchp-nocheck-1     State.Status=running  Health=None
Name=nchp-failcheck-1   State.Status=running  Health={'Status': 'unhealthy', 'FailingStreak': 17}
```

and exactly two timers, one per container that declares a check:

```
$ systemctl list-units --all --type=timer | grep nerdctl
6e7d785b...timer  loaded active waiting  nerdctl --namespace=nchp container healthcheck 6e7d785b...
aed39a59...timer  loaded active waiting  nerdctl --namespace=nchp container healthcheck aed39a59...
```

Two details the run corrected, both now pinned by
`real_nerdctl_inspect_output_parses` against the captured payload rather than a
hand-written one:

- `Name` has **no** leading slash (`nchp-failcheck-1`, where Docker gives
  `/nchp-failcheck-1`). Handled, but it was an assumption before.
- A container with no healthcheck has **no `Health` object at all** — not
  `"Status": "none"`. That is precisely the case `judge()` treats as "not
  judged", so it mattered that it was right.

*3. And it all works inside a real TDX CVM with a read-only rootfs.*

The remaining worry was that the guest rootfs is immutable, so a transient
systemd unit might have nowhere to land. Booted a dev guest image
(`dstack-mkosi-dev-496128cb0`, nerdctl 2.2.1) as a real TDX CVM via
`dstack-vmm run`, with the probe as the app-compose `bash_script`:

```
--- rootfs mount flags ---
/dev/mapper/rootfs squashfs ro,relatime,errors=continue
--- /var/lib overlay ---
overlay overlay rw,relatime,lowerdir=/var/lib,upperdir=/var/volatile/lib/upper,workdir=/var/volatile/lib/work
dstack/data[/var/lib/nerdctl] zfs rw,noatime,xattr,noacl,casesensitive

===== TEST 1: can systemd create a transient timer at all? =====
systemd-run rc=0
ActiveState=active
FragmentPath=/run/systemd/transient/probe-timer.timer
Transient=yes

===== TEST 2: nerdctl healthcheck timer =====
nerdctl version 2.2.1
--- timers created ---
  bf01d716...timer  loaded active waiting  [systemd-run] /usr/bin/nerdctl \
      --namespace=dstack container healthcheck bf01d716...
--- inspect health ---
"Health": { "Status": "healthy", "FailingStreak": 0, "Log": [ ... ] }
```

The root filesystem really is `ro` (dm-verity squashfs), and it does not matter:
the timer is transient and lands in `/run/systemd/transient/` (tmpfs), while
nerdctl's own state is writable because `dstack-volatile-binds.sh` overlays
`/var/lib` with a tmpfs upper and `dstack-prepare.sh` then binds the encrypted
data disk over `/var/lib/nerdctl` and `/var/lib/containerd`.

Worth noting the timer fired and the check reported `healthy` on **nerdctl
2.2.1** — the version currently shipped. The timer machinery is not what needs
2.3.1; only the Compose-layer translation of `healthcheck:` is.

(All probes cleaned up: the CVM was shut down, the scratch workdir removed, and
on the host containerd namespaces are back to `moby`/`moby_history` with no
leftover units.)


kvinwang:
Folded into #1079 — the whole health story now lives in one PR (guest-agent `Worker.Health`, gateway polling and gating, nerdctl support, the mkosi nerdctl 2.3.5 bump). Nothing is dropped; the commits were rebased across unchanged.

PR #1082, opened 2026-08-20
feat(app-compose): let an app declare its own health check

kvinwang:
> Stacked on #1080.

## Problem

Health can only be derived from containers that declare a Compose
`healthcheck`. That covers `docker-compose` and `nerdctl-compose` and leaves out
everything else:

- The **`bash` runner** has no containers to inspect. #1080 gives it
  `not_judged()` — always healthy — which is the only safe default but means the
  gate does nothing for it.
- **Any runner added later** starts in the same position, with no way to
  participate short of another patch to the guest agent.
- A Compose app that wants to answer for the application **as a whole**, rather
  than have every container judged separately, has no way to say so.

## Fix

An optional `health_check` in `app-compose.json`. The guest agent runs the named
program on a timer and caches the verdict; `Worker.Health` serves the cache.

```json
{
  "runner": "bash",
  "bash_script": "/opt/my-app/run.sh &",
  "health_check": {
    "path": "/opt/my-app/healthz.sh",
    "args": ["--strict"],
    "interval_secs": 10,
    "timeout_secs": 5
  }
}
```

Exit 0 is healthy; anything else — non-zero, timeout, or a program that cannot
be executed — is not.

**Run on a local timer, not on each gateway poll.** Several gateway nodes poll
the same instance, and the poll interval is an operator's setting on the
gateway; neither should decide how often the app's own probe executes. The two
cadences stay independent, and the cached verdict is what gets served.

**Declared means it decides, for every runner.** It overrides container-derived
health rather than combining with it — one answer from the app beats an
inference assembled from container state, and a single source avoids the
question of what to do when the two disagree.

**Not having run yet is not a pass.** The probe starts out reporting unhealthy
with `health check has not run yet`, the same rule the gateway applies to an
instance that has registered but not answered a poll. Registration happens long
before the app is up.

**`path` is a program, not a shell command line.** Nothing in it is word-split,
glob-expanded or variable-substituted, and arguments go in `args`. If you want
shell semantics, point it at a script.

**Output is bounded.** The exit code plus up to 512 bytes of stderr reach the
gateway's logs; a probe that dumps a stack trace or an HTML error page cannot
flood them.

### On the trust boundary

The probe runs as root inside the CVM. That is not new privilege: `app-compose`
is measured into `compose_hash` and attested, the CVM is single-tenant, and the
app already runs root code through `bash_script` / `init_script`. This adds a
place to *declare* a command, not a place to smuggle one in.

### Compatibility

`compose_hash` is `sha256_file(app-compose.json)` — over the raw file bytes, not
a re-serialisation — so adding an optional field changes no existing app's hash.
Absent `health_check` behaves exactly as #1080 does today.

## Documentation

`docs/app-health-checks.md` is new, and covers the whole picture across #1078 –
#1081 rather than just this field: where the verdict comes from and in what
order, what container healthchecks do and do not judge, the nerdctl >= 2.3.1
requirement, how the gateway uses the result, why health fails open while the
operator gate does not, and the gateway-side config. There was no user-facing
documentation of any of this before.

## Verification

`cargo test -p dstack-guest-agent`: 48 passed (7 new).
`cargo test -p dstack-util --all-features`: 89 passed.
`cargo test -p dstack-gateway --all-features`: 192 passed.
`cargo fmt --check` and `clippy -D warnings` clean across guest-agent, dstack-util and gateway.

New tests drive real subprocesses rather than a mock, one per behaviour above:

- `exit_zero_is_healthy`
- `a_non_zero_exit_reports_the_code`
- `stderr_is_carried_into_the_reason`
- `a_hanging_probe_times_out_as_unhealthy`
- `a_missing_program_is_unhealthy_rather_than_an_error`
- `a_noisy_probe_has_its_output_bounded`
- `the_first_report_is_not_healthy`

## Follow-up, not in this PR

`vmm-cli.py` hardcodes `runner: "docker-compose"` and exposes no `runner` flag,
so it cannot create a `bash`-runner app at all today. Adding `--health-check`
there is worth doing once it can, but it is a separate change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - gateway: `Admin.SetInstanceReady` takes a CVM instance out of its app's load-balancing rotation without stopping it; instance-id routing stays open so the instance can still be investigated, and the setting survives re-registration
 - gateway: application-level health polling. The gateway asks each CVM's guest agent (new `Worker.Health` RPC) whether every container declaring a Compose `healthcheck` is healthy, and keeps instances that say no -- or that have not answered since registering -- out of app-id load balancing. Legacy images, apps without healthchecks, and apps where no instance reports healthy are unaffected
 - guest-agent: container health also covers the `nerdctl-compose` runner, read through `nerdctl inspect` (its output is Docker-compatible). Requires nerdctl >= 2.3.1 for Compose `healthcheck:` to be honoured; the mkosi backend is pinned to 2.3.5
+- app-compose: optional `health_check` (program path, args, interval, timeout) run periodically by the guest agent. Overrides container-derived health for any runner, and is the only way the `bash` runner — or any future non-container runner — can report health. Documented in `docs/app-health-checks.md`
 
 ## [0.5.5] - 2025-10-20
 
--- a/docs/app-health-checks.md
+++ b/docs/app-health-checks.md
@@ -1,0 +1,133 @@
+# Application health checks
+
+The gateway keeps an instance out of its app's load-balancing rotation while the
+application is not able to serve. This page describes where that verdict comes
+from and how an app influences it.
+
+## Why this exists
+
+A CVM registers with the gateway from `dstack-util setup`, during boot.
+`app-compose.service` is ordered *after* `dstack-prepare.service`, so at the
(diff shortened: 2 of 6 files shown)


kvinwang:
Folded into #1079 for the same reason as #1080 — one PR for the whole health story. The app-declared `health_check` field and `docs/app-health-checks.md` are both there, unchanged.

PR #1083, opened 2026-08-20
build(os/yocto): nerdctl 2.3.5 for compose healthcheck support

kvinwang:
## Why

The Yocto backend builds nerdctl from `release/2.2`, and `nerdctl compose` did
not translate the Compose `healthcheck:` field into `nerdctl run --health-*`
flags until **v2.3.1**. Without it, the health gating added in #1079 has nothing
to read on the `nerdctl-compose` runner: containers are visible, but none of them
ever has a healthcheck configured.

Yocto is the default backend (`os/build.sh`), so this is what carries the feature
to production images. The mkosi backend was already moved to 2.3.5 in #1079.

## What

Two submodule pointer bumps.

**`os/yocto/deps/meta-virtualization`** → `fce97db68e` — the nerdctl 2.3.5 recipe
bump in Dstack-TEE/meta-virtualization#1: `SRCREV`, `branch=release/2.3`,
`PV = v2.3.5`, plus the five regenerated Go module manifests.

**`os/yocto/deps/openembedded-core`**: `06dd66e622` → `ef022bf82d` (`wrynose`
head). This is not optional garnish — nerdctl's `go.mod` `go` directive is
`1.26.3` from v2.3.1 onwards, the pinned OE-core provides only `go 1.26.2`, and
both nerdctl's `do_compile` and `go-mod-discovery.bbclass` export
`GOTOOLCHAIN="local"`. Go therefore hard-refuses to build it. `wrynose` head
ships `go_1.26.5.bb`.

## What the OE-core move actually drags in

Worth being explicit, because it is larger than "a Go bump": **347 files changed,
+18122/-23675**.

| | |
|---|---|
| **gcc 15.2 → 15.3** | the headline — rebuilds everything, so every image measurement changes |
| go 1.26.2 → 1.26.5 | the reason for the move |
| openssl 3.5.6 → 3.5.7, glib 2.88.0 → 2.88.2, util-linux 2.41.3 → 2.41.5, bind 9.20.22 → 9.20.26, socat 1.8.1.1 → 1.8.1.3 | routine `wrynose` backports |

Checked for breakage against our layers:

- **No recipe deletions.** Every `.bb` removal is a rename (version bump); the
  only pure deletions are patches that went upstream, plus
  `cve-exclusion_6.18.inc`, which was consolidated into `cve-exclusion.inc`
  inside OE-core and is `require`d by `linux-yocto` itself, not by us.
- **`linux-yocto_6.18.bb` still exists**, so
  `PREFERRED_VERSION_linux-yocto ?= "6.18%"` in `meta-dstack/conf/distro/dstack.conf`
  still resolves.
- **`BB_MIN_VERSION` is `2.18.0` before and after**, so the pinned
  `bitbake` 2.18 submodule needs no bump, and neither does `meta-yocto`.
- **No version-pinned references** to gcc 15.2, util-linux 2.41.3, openssl 3.5.6
  or go 1.26.2 anywhere in `os/yocto/layers/`.
- All six Go components in the image — `docker-moby`, `docker-compose`,
  `nerdctl`, `containerd-cni`, `stargz-snapshotter`, `dstack-sysbox` — move to
  go 1.26.5. From the `go` directive's point of view this is permissive:
  `GOTOOLCHAIN=local` only fails when the toolchain is *older* than the module
  declares, so a newer toolchain can unblock but not block. The residual risk is
  behavioural compiler change, not refusal to build.

The nerdctl recipe's `do_compile` / `do_install` are unchanged, deliberately: the
`binaries` Makefile target, `GO_BUILD_FLAGS` / `BUILDTAGS` and the
`_output/nerdctl` path are identical between v2.2.1 and v2.3.5.

## Release impact

gcc 15.2 → 15.3 rebuilds the world, so **MRTD and RTMR values change for every
guest image**. The new image hashes have to be added to the KMS on-chain
whitelist before deployments can use them. This is a release event, not just a
code change.

## Verification

Built from exactly these two pointers with `bitbake nerdctl`:

```
NOTE: recipe nerdctl-v2.3.5-r0: task do_fetch: Succeeded
NOTE: recipe nerdctl-v2.3.5-r0: task do_compile: Succeeded
NOTE: recipe nerdctl-v2.3.5-r0: task do_install: Succeeded
NOTE: Tasks Summary: Attempted 1955 tasks of which 1935 didn't need to be rerun and all succeeded.
```

The resulting binary is a 35 MB x86-64 ELF whose embedded toolchain string is
`go1.26.5` — i.e. the OE-core move is doing exactly the job it was made for, and
nerdctl v2.3.5 compiles against it.

The first attempt **failed**, which is the reason this was worth building rather
than reviewing:

```
ERROR: nerdctl-v2.3.5-r0 do_fetch: Missing SRC_URI checksum
```

`oe-go-mod-fetcher-hybrid.py` downloads each module's zip to compute its
sha256, and one of them hit a TLS handshake timeout against proxy.golang.org.
The script logged a warning, emitted the `gomod://` line **without**
`sha256sum=`, and exited 0 — a transient network fault turning into a silently
incomplete manifest. Re-running just moved the failure to a different random
module, since there is no retry. The checksums were instead filled in
deterministically (they are the sha256 of the module zip from the proxy,
verified against one the fetcher had got right) and every entry was then
checked: 610 `gomod://` entries all carry `sha256sum=`, 597 + 36 `git://`
entries all carry `rev=`, no truncated or empty values in any of the five files.

**CI does not cover any of this.** `guest-os.yml` — the Yocto build — is
`workflow_dispatch` only, so it never runs on a pull request. What the `os/**`
path filter does trigger is `mkosi-build.yml`, which builds the *other* backend
and is untouched by these two submodule pointers. So the checks going green on
this PR say nothing about the change in it; the `bitbake nerdctl` run above is
the only evidence, and it covers one recipe rather than the whole image.

A full `make os-image` from these pointers is the honest test, since gcc 15.3
rebuilds every recipe and any one of them could be the one that breaks. That is
running now and I will post the result.

Worth knowing for CI planning: the gcc 15.2 → 15.3 move invalidates essentially
the whole sstate cache, so the first build after this merges is a world rebuild
rather than an incremental one — 1955 tasks for `nerdctl` alone, 6778 for
`dstack-rootfs`. It is also memory-hungry: a full-parallelism run on a 64-core /
125 GB host (shared with other work) exhausted swap and the bitbake server was
OOM-killed two thirds of the way through. Re-running it capped at
`BB_NUMBER_THREADS=10 PARALLEL_MAKE=-j6` is fine. Worth checking the
`yocto-builder` runner has headroom before the first post-merge build.

## Status

- [x] `openembedded-core` pointer bumped
- [x] `meta-virtualization` pointer bumped to `fce97db68e`, now the merged head
      of `dstack-wrynose` (Dstack-TEE/meta-virtualization#1 was fast-forwarded,
      so the branch commit *is* the merge commit — no re-pointing needed)
- [x] `nerdctl` builds from this tree (see above)
- [ ] Full guest image rebuilt and booted
- [ ] New image hashes whitelisted in KMS

A `bitbake nerdctl` run from exactly these two pointers is in progress. What it
has already established, from the parse phase:

- The five regenerated manifests parse cleanly — **0 errors** — so the SRC_URI
  entries are well formed and the layer set is self-consistent.
- The build plan resolves `go-binary-native-1.26.5`, `go-cross_1.26.5` and
  `go-cross-x86-64-v3-1.26.5`, i.e. the OE-core move does supply the toolchain
  nerdctl v2.3.5 requires.

Still pending: whether nerdctl itself compiles under go 1.26.5, which is the last
task in the graph.

Worth knowing for CI planning: the gcc 15.2 → 15.3 move invalidates essentially
the whole sstate cache, so this verification is a 1955-task world rebuild rather
than an incremental one. The same cost lands on the first CI run and the first
image build after this merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - shared API authentication (`dstack-api-auth`) protecting the full VMM HTTP/pRPC/UI surface and unifying Gateway/KMS admin auth: bearer/`X-Admin-Token`/HTTP Basic/bcrypt htpasswd, constant-time verification (#796)
 
+### Changed
+- os/yocto: nerdctl 2.2.1 → 2.3.5, so `nerdctl compose` honours the Compose `healthcheck:` field (only translated into `--health-*` flags from 2.3.1 on). Requires openembedded-core to move to `wrynose` head for go 1.26.5, which also brings gcc 15.2 → 15.3 — every guest image measurement changes, so the new image hashes need whitelisting in KMS
+
 ## [0.5.5] - 2025-10-20
 
 ### Added


kvinwang:
Post-merge follow-up: the full guest image now builds from these pointers, not
just `nerdctl` on its own.

```
NOTE: Tasks Summary: Attempted 6778 tasks of which 6732 didn't need to be rerun and all succeeded.
```

```
$ grep '^nerdctl ' dstack-rootfs-*.manifest
nerdctl x86_64_v3 v2.3.5

$ ls -la deploy/images/*/dstack-rootfs-*
... 574783488  dstack-rootfs-...squashfs
... 1571147776 dstack-rootfs-...cpio
```

So gcc 15.2 → 15.3 breaks nothing across the tree — kernel, systemd, docker,
containerd, nerdctl, zfs, sysbox and clang all rebuild — and v2.3.5 is what
lands in the image. Worth having on record because `guest-os.yml` is
`workflow_dispatch` only, so no CI run covers any of this.

Two things that came out of running it, both about the build host rather than
this change:

- At full parallelism on a 64-core / 125 GB host shared with other work, swap
  was exhausted and the bitbake server was OOM-killed two thirds of the way in.
  `BB_NUMBER_THREADS=10 PARALLEL_MAKE=-j6` completed without trouble. Worth
  checking the `yocto-builder` runner has headroom, since the gcc bump makes the
  first post-merge build a world rebuild.
- `SOURCE_MIRROR_URL = "https://mirrors.kernel.org/yocto-sources/"` (set in the
  generated `bb-build/conf/local.conf`, not in the repo) was accepting
  connections and then sending nothing, hanging every fetch behind it —
  `docker-compose` alone pulls hundreds of Go modules through it. Dropping
  `own-mirrors` for the run let it finish immediately. Not a repo problem, but
  if CI ever inherits that mirror setting it will look like a hung build.

⚠️ Separately: `next` currently records a `meta-virtualization` pointer that does
not exist, which came from this PR. #1085 fixes it. This build passing did *not*
catch that — the local submodule already had the correct commit checked out, so
the build validated the working tree rather than the recorded pointer.

kvinwang:
Follow-up to the build result above: the image also **boots**.

Assembled the full artifact set (`bitbake virtual/kernel dstack-initramfs
dstack-ovmf dstack-guest`, then `mc:prod:dstack-rootfs` + `mc:prod:dstack-uki`,
then `mkimage.sh`) and booted it as a real TDX CVM via `dstack-vmm run`:

```
kernel:       6.18.39-dstack
systemd:      starting
app-compose:  ran the probe as root
failed units: (none)
Kernel panic: 0
```

`/usr/bin/nerdctl`, `/usr/bin/containerd` and `/usr/bin/docker` are all present.

Zero failed systemd units is the signal that matters here — a toolchain bump
that produced a subtly broken userland would normally show up as services
failing to start, and nothing did.

This closes the gap between "the tree compiles under gcc 15.3" and "the image
built from it actually runs", which the build result alone did not cover.

One incidental: assembling the image needs `nitro-tpm-pcr-compute` on the host
for `measurement.aws.cbor`. It is not vendored, and `mkimage.sh` fails without
it. The error message names the pinned rev to `cargo install`, so this is only
a note for whoever runs the first post-merge image build.

PR #1089, opened 2026-08-20
fix(gateway): keep KV record fields the writing build does not declare

kvinwang:
## Problem

KV values are msgpack maps keyed by field name, so a gateway running an older build
already *reads* a record written by a newer one without tripping over the fields it does
not know. `kv/mod.rs` documents exactly that, and it is only half of a rolling upgrade.

The other half is the write path. `encode` serializes the struct as the running binary
declares it, so the encoding contains those fields and nothing else. The moment an older
node rewrites a record, every field added by a later release is gone — for the whole
cluster, since the record then replicates as the newest write. The node that added the
field has no way to notice.

This is not a read-modify-write bug that a `update()` helper would fix, because the worst
path never reads the KV type at all. `inst/` records are hydrated into `InstanceInfo`
(`main_service.rs`), a hand-mirrored in-memory struct, and every write-back rebuilds
`InstanceData` field by field (`main_service.rs:1028`, `:1138`, `:1155`). An unknown field
is dropped at hydration; the next re-registration, admin call or lazily-filled port policy
persists that loss. `update_dns_credential` and `set_certbot_config` are true
read-modify-writes with the same outcome.

Concretely: node A on a newer build writes `inst/x` with a new `health_probe_path`; node B
on the current build re-registers the same instance; the field is gone everywhere, and A
keeps serving as though it had never been set.

## Fix

`kv/compat.rs` carries forward, at write time, the fields present in the stored record that
the running binary does not declare. `put_encoded` splices them onto the outgoing encoding
before the `put`.

Three decisions worth reviewing:

**Unknown is decided against the field list `T` declares, not against the keys missing from
the new encoding.** The list comes from the derived `Deserialize` impl, via a `Deserializer`
whose only job is to capture what `deserialize_struct` hands it. This matters because
`#[serde(skip_serializing_if)]` — `DnsProvider::Cloudflare::api_url` uses it — makes a field
this build deliberately *cleared* absent from the encoding for exactly the same reason a
field from the future is. Keying on the declaration keeps the clear and still carries the
unknown. Aliases and renames count as declared, so `#[serde(alias)]` is safe too. The probe
answers `None` for scalars, `#[serde(flatten)]` types and internally tagged enums, and the
merge then declines entirely — that last one on purpose, since an internally tagged enum's
variants have disjoint fields and merging across a variant change would graft one variant's
secrets onto another.

**Only long-lived records carry fields; snapshots do not.** `inst/`, `node/info/`,
`dns_cred/`, `cert/*/config` and `global/certbot_config` are records that a write updates.
A certificate, an attestation, a lock or a counter is a complete new fact on every write,
and carrying a field across one would attribute the previous snapshot's value to the new
one — an older gateway renewing a certificate would publish a stale chain next to the key
it just issued. Absent is a case the newer reader already has to handle; stale is not.

**Only the root map is merged.** Below it nothing distinguishes a struct encoded as a map
from a `BTreeMap` field, so recursing would resurrect map entries that were deliberately
removed. New fields therefore go at the root of a value type; a nested type that has to
grow needs its own `#[serde(flatten)]` catch-all. A field this build stops declaring is
carried forever, since it is indistinguishable from one a newer peer wrote — retiring a
field for real needs an explicit drop list, which is left for the release that first needs
it rather than landing empty here.

The merge is byte-level: the header is rewritten and the known fields keep the exact bytes
`rmp_serde` produced, so nothing is re-encoded and unknown values are copied verbatim,
whatever their type. It uses `rmp::Marker` — already in the tree under `rmp-serde`, not a
new supply-chain entry — rather than a hand-copied opcode table, and the walker is
iterative so a deeply nested value cannot exhaust the stack. Every failure declines the
merge and writes the original encoding, which is today's behavior.

## Verification

`cargo test -p dstack-gateway`: 184 passed. `cargo clippy -p dstack-gateway --all-targets
-- -D warnings`: clean.

End-to-end through a real `KvStore` (`value_encoding_tests`): a record written with an
extra `health_probe_path`, then re-registered by this build via `sync_instance`, still has
the field and still decodes as `InstanceData`; the same setup on `cert/{domain}/data`
confirms a snapshot does *not* inherit the previous write's `chain_pem`.

The byte walker is differential-tested against `rmpv` (dev-dependency only):

| Test | Cases | Result |
| --- | --- | --- |
| Random values, every header width (str/bin/ext/array/map, 4 levels deep) | 2,000 | Lands on the same byte as `rmpv` |
| Valid encodings with 1–3 bit flips and random truncation | 50,000 | Same accept/reject and same length |
| Arbitrary random bytes | 100,000 | No panic, never reads past the buffer |
| 200,000-deep nesting | 1 | Parsed, no stack growth |

The one divergence found across the fuzz corpus is the reserved marker `0xc1`, which `rmpv`
decodes as `Nil` and this walker rejects. `rmp_serde` also rejects it
(`Err(TypeMismatch(Reserved))`), so the walker matches the decoder that actually reads these
values; the test asserts the exception rather than hiding it.

## Rollout note

Like widening a key schema, this only helps once it is deployed: a field added in release
N+1 survives only if the nodes running release N already carry unknown fields. This lands
before any further field is added to a KV value type, not with it.

## Upgrade note

This changes a durability default. `sync.wal_sync_interval` ships at `"5s"`, so after an
upgrade a KV write is no longer forced to the disk before it returns; it is forced within the
window instead. Losing the process costs nothing either way — the bytes reach the kernel on
every write, so a panic, an OOM kill or a restart loses none of them, which
`test_wal_integrity` demonstrates by SIGKILLing a gateway two seconds after five
registrations and finding all of them in the log. Losing the *machine* costs up to the
window.

A cluster recovers that window from its peers: acks never become durable ahead of the data
they cover, so a restarted node cannot claim coverage of what it lost. A single-node gateway
has no such backstop, and holds the only copy of its ACME account and DNS credentials.
Operators running one should set `wal_sync_interval = "0s"`, which restores the previous
behaviour exactly.

### DIFF
--- a/dstack/Cargo.lock
+++ b/dstack/Cargo.lock
@@ -1925,7 +1925,7 @@ dependencies = [
  "tracing-subscriber",
  "uuid",
  "wavekv 1.0.0",
- "wavekv 2.0.0",
+ "wavekv 2.1.0",
  "x509-parser",
 ]
 
@@ -8406,9 +8406,9 @@ dependencies = [
 
 [[package]]
 name = "wavekv"
-version = "2.0.0"
+version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1c12ef936041e6aea20bacac4ff79b7c93adc57b232e9d381ecd8e2b1c5f4753"
+checksum = "99f5f24dfab68f43806c0a15c3b87d9cde250dfc876cae3ec34642427becf336"
 dependencies = [
  "anyhow",
  "bincode 2.0.1",
--- a/dstack/Cargo.lock
+++ b/dstack/Cargo.lock
@@ -1904,7 +1904,9 @@ dependencies = [
  "rand 0.8.6",
  "reqwest",
  "rinja",
+ "rmp",
  "rmp-serde",
+ "rmpv",
  "rocket",
  "rustls",
  "safe-write",
@@ -2584,7 +2586,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "39cab71617ae0d63f51a36d69f866391735b51691dbda63cf6f96d042b63efeb"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -3478,7 +3480,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "tokio",
  "tower-service",
  "tracing",
@@ -3808,7 +3810,7 @@ checksum = "3640c1c38b8e4e43584d8df18be5fc6b0aa314ce6ebf51b53313d4306cca8e46"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -4924,7 +4926,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7d8fae84b431384b68627d0f9b3b1245fcf9f46f6c0e3dc902e9dce64edd1967"
 dependencies = [
  "libc",
- "windows-sys 0.48.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -5610,7 +5612,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
@@ -5649,9 +5651,9 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.5.10",
+ "socket2 0.6.4",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -6124,6 +6126,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rmpv"
+version = "1.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7a4e1d4b9b938a26d2996af33229f0ca0956c652c1375067f0b45291c1df8417"
+dependencies = [
+ "rmp",
+]
+
+[[package]]
 name = "rocket"
 version = "0.6.0-dev"
 source = "git+https://github.com/rwf2/Rocket?branch=master#3a54d079aef060a8f732bd04ea54b0581a604087"
@@ -6303,7 +6314,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.12.1",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -6371,7 +6382,7 @@ dependencies = [
  "security-framework",
  "security-framework-sys",
  "webpki-root-certs",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -7542,7 +7553,7 @@ dependencies = [
  "getrandom 0.4.2",
  "once_cell",
  "rustix 1.1.4",
- "windows-sys 0.52.0",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


kvinwang:
Follow-up in `bba4773093`: the declared-field list says which names `T` knows, not what it tolerates beside them. A `deny_unknown_fields` type (none today, but nothing stops one) would have taken the carried field and become undecodable to its own writer — strictly worse than the field being dropped. The merged bytes are now decoded as `T` before they are written, and the merge is abandoned if that fails; the check only runs on the carry path, which is the version-skew window, not the steady state.

Also added tests for the shapes that look like maps on the wire but are not structs — a `BTreeMap` value (first byte 0x81) and an externally tagged enum (0x81) — where carrying a "missing" key would resurrect a deliberately removed entry or graft one variant's fields onto another. The byte walker cannot tell them from a struct; the declared-field probe is what refuses them.

kvinwang:
Measured the write path rather than asserting it, since "free in the steady state" was too strong on my part — the round-trip guard is on the carry path only, but the walk that decides there is nothing to carry is not.

Per `put_encoded`, 200k iterations, release build, `InstanceData` at 346 bytes:

| Step | Cost |
| --- | --- |
| `is_long_lived_record` (all a non-record key pays) | 2.6 ns |
| `get` + value clone | 76 ns |
| Merge, nothing to carry | 465 ns |
| Merge, carrying one field (incl. round-trip decode) | 2,166 ns |

So a record-key write costs ~545 ns more than before, against a wavekv persistent `put` of 165–184 µs — the WAL write dominates — which puts it at ~0.3%. A non-record key (`conn/`, `handshake/`, `last_seen/`, `node/status/`, `__peer_addr/`) pays the 2.6 ns prefix test and nothing else: it never reaches the `get`.

`49d74dc7de` halves the steady-state figure — 1,025 ns to 465 ns — by settling the question before parsing the outgoing encoding. In a single-version cluster the stored record holds nothing unknown, so the common write now walks one map instead of two. The result is identical either way: the early exit only triggers when the carry set would have been empty.

kvinwang:
Replaced the key-prefix classification with a parameter on `put_encoded` in `7a611ae151`.

The old `is_long_lived_record(key)` decided, by string-matching a key in `compat.rs`, something the calling code already knew: `save_cert_data` is aware that it is writing a whole new certificate. Splitting the knowledge from the decision made it a cross-cutting table that can drift from its call sites, and its failure mode was silent — a key nobody added to it simply lost the protection this module exists to provide.

`put_encoded` now takes `UnknownFields::Keep` or `UnknownFields::Drop`. `Keep` for a write that updates a record outliving it — instance, node, DNS credential, ZT-domain config, certbot config. `Drop` for one that states a complete new fact — a certificate, an attestation, a lock, a counter, a peer URL. Behaviour is identical to the previous commit at every existing site; what changes is that the reasoning now sits where the write happens, and a new write does not compile until it says which it is.

The probe still runs before the stored record is read, so a `Drop` write and a non-struct value both cost a branch rather than a lookup.

`cargo test -p dstack-gateway`: 185 pass (the classification test went with the classification). Clippy clean. The two end-to-end tests are unchanged and now assert the call sites' own choices: `sync_instance` keeps a field it cannot see, `save_cert_data` does not inherit the previous certificate's.

kvinwang:
Wired the WaveKV 2.1 window in `1dedf15310`, so the gateway actually gets the write-path half of it.

`sync.wal_sync_interval` — "5s" by default — reaches `KvStore::new` and from there the persistent node's `NodeConfig`; the ephemeral store keeps no log and is untouched. A task next to the existing persist loop calls `sync_wal_if_due()` on that same interval, because nothing else forces the log on a schedule: a snapshot does, but `persist_interval` is five minutes. Ticking at the window is enough to make it a bound rather than a hope — the store measures from its last fsync, so a write is forced at the first tick that finds the window elapsed and never waits more than one window. Setting it to "0s" restores exactly the old behaviour, every write forced before it returns, which is what a single-node gateway holding the only copy of its ACME account should keep.

What the window costs is bounded to losing the *machine*. Losing the process — a panic, an OOM kill, a restart — costs nothing either way, because `append_ops` still flushes to the kernel on every write and only the fsync is deferred. In a cluster even the machine case is recoverable: acks never become durable ahead of the data they cover, so a restarted node cannot claim coverage of what it lost and its peers re-offer. That reasoning is in the `gateway.toml` comment where an operator will actually read it, not just here.

No shutdown hook: an unclean kill loses at most the window, which is the number the operator already agreed to.

Two tests pin the wiring rather than the library's behaviour — that the interval reaches the store at all, and that `None` still fsyncs per write. The first fails if the config stops at `SyncConfig`, which is the way this would plausibly break.

`cargo test -p dstack-gateway`: 187 pass. Clippy clean.

PR #1090, opened 2026-08-21
test(gateway): register with public keys the gateway can accept

kvinwang:
## Problem

Ten of the twenty-eight cases in `gateway/test-run/test_suite.sh` cannot pass. Every one of
them registers a CVM with a hand-written placeholder public key, and every one of those keys
is the wrong length:

```
"error": "[cafebabe] invalid client public key: public key is 41 characters, expected 44"
```

A WireGuard public key is 32 bytes, so 44 base64 characters. The thirteen placeholders in the
suite are 38 to 43:

| Placeholder | Length |
| --- | --- |
| `bootnode12345678901234567890123456789=` | 38 |
| `threenode12345678901234567890123456789=` | 38 |
| `stresstest%02d12345678901234567890123456=` | 39 |
| `partition%02d123456789012345678901234567=` | 39 |
| `backup123456789012345678901234567890123=` | 40 |
| `persist%02d123456789012345678901234567890=` | 40 |
| `upnode…`, `uptest…` | 40 |
| `prpctest…`, `waltest%02d…`, `downtest…`, `downnode…` | 41 |
| `persisttest1234567890123456789012345678901=` | 43 |

They were fine until registration started validating the key length
(`WG_PUBLIC_KEY_B64_LEN` in `gateway/src/kv/import.rs`, added with the recovery-API review
follow-ups). The suite is not in CI — CI runs `test_proxy.sh` — so nothing said so.

The cases this silences are the ones that cover persistence and replication most directly:
`test_wal_integrity`, `test_periodic_persistence`, `test_client_registration_persistence`,
`test_stress_writes`, `test_network_partition`, `test_three_node_cluster`,
`test_three_node_bootnode`, `test_prpc_register`, and both node-status registration cases.
`test_periodic_persistence` never reaches its persistence assertion at all.

## Fix

Each placeholder becomes a call to `test_public_key`, the generator the suite already has and
that the other five call sites already use. It emits a real 32-byte key, so the gateway
accepts it, and it is deterministic, so a failure is still reproducible. Seeds are distinct
per registration — two instances sharing a public key are refused as a conflict, which would
just be the same class of bug again.

The generator's doc comment now says why a placeholder is not a shortcut, so the next one
does not get written.

## Verification

Before, on `next`: 18 passed, 10 failed — the ten above.

After, each previously-failing case run individually:

```
test_prpc_register                       PASSED
test_client_registration_persistence     PASSED
test_stress_writes                       PASSED
test_network_partition                   PASSED
test_wal_integrity                       PASSED   (Registered 5/5 clients, WAL 944 bytes)
```

with the rest still running at the time of writing; I will update this if any of them turn
out to be failing for a second, unrelated reason.

`shellcheck -S error` clean, `bash -n` clean. No production code is touched.

### DIFF
--- a/dstack/gateway/test-run/test_suite.sh
+++ b/dstack/gateway/test-run/test_suite.sh
@@ -412,6 +412,11 @@ get_node_uuid() {
 	admin_get_status "$admin_port" | python3 -c "import sys,json; print(json.dumps(json.load(sys.stdin)['uuid']))" 2>/dev/null || true
 }
 
+# A deterministic, *valid* WireGuard public key: 32 bytes, so 44 base64
+# characters. Registration rejects anything else (see WG_PUBLIC_KEY_B64_LEN in
+# gateway/src/kv/import.rs), so a hand-written placeholder is not a shortcut —
+# it is a test that can only fail. Use a distinct seed per registration; two
+# instances sharing a public key are refused as a conflict.
 test_public_key() {
 	local seed=$1
 	python3 -c "import base64; print(base64.b64encode(int($seed).to_bytes(32, 'big')).decode())"
@@ -1035,7 +1040,7 @@ test_prpc_register() {
 	log_info "Debug service is available"
 
 	# Register via debug port
-	local register_response=$(debug_register_cvm $debug_port "prpctest12345678901234567890123456789012=" "deadbeef" "cafebabe")
+	local register_response=$(debug_register_cvm $debug_port "$(test_public_key 501)" "deadbeef" "cafebabe")
 	log_info "Register response: $register_response"
 
 	# Verify registration succeeded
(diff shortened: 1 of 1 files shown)


kvinwang:
All ten now pass, run individually against a release build of `next` + this change:

```
test_prpc_register                       PASSED
test_client_registration_persistence     PASSED
test_stress_writes                       PASSED
test_network_partition                   PASSED
test_three_node_cluster                  PASSED
test_three_node_bootnode                 PASSED
test_periodic_persistence                PASSED
test_node_status_register_exclude        PASSED
test_node_status_register_reject         PASSED
test_wal_integrity                       PASSED   (Registered 5/5, WAL 944 bytes)
```

None of them turned out to have a second, unrelated reason to fail — the key length was the
only thing standing in front of all ten.

PR #360, opened 2025-09-29
Add Verifiable Message Signing

iKapitonau:

PR #1110, opened 2026-08-24
refactor!: move signature verification out of the CVM and into the SDKs

kvinwang:
## What

Removes the guest-agent `Verify` RPC and moves signature verification into the four SDKs, where it also gains the check it was always missing: walking the `signature_chain` back to a KMS root key the caller independently trusts.

## Why `Verify` should never have been an RPC

Checking a signature needs no key material and no attestation. The agent's verdict also came back over the socket **unattested** — so a caller who believed the TEE was vouching for it was mistaken, and a caller who did not gained nothing over checking the signature themselves. Meanwhile it parsed attacker-supplied keys and signatures inside the TEE on every call: inbound attack surface for a non-benefit.

`Sign` stays server-side. It needs a key only the TEE holds.

The RPC only existed because review #360 asked for a `Sign()` counterpart. That parity requirement is real, but it is satisfied by an SDK-local function — it never needed to be an endpoint.

## ⚠️ Breaking change

`Verify` shipped in **v0.5.6 through v0.5.9**. An SDK pinned at 0.5.x calling `/Verify` against a 0.6+ guest agent now gets an unknown-method error.

Note that `sdk/curl/api.md` and `sdk/rust/README.md` both labelled Sign/Verify *"not yet released"*. That annotation was **stale** — this PR removes it. It would have made this look like a free cleanup when it is not.

The client-side `verify()` method is also gone from all four SDKs, replaced by a standalone function. A pure function of four byte strings had no business hanging off an object that holds a socket connection. **Happy to add thin deprecated delegates back if you would rather stage that separately.**

## New capability: `verify_signature_chain`

The old RPC checked one signature against a public key the caller passed in — that proves only that *whoever holds that key* signed the data, not whose key it is. The chain verifier walks all three links:

| Link | Signed message | Format | Hash |
|---|---|---|---|
| `[0]` payload | the payload | 64B `r‖s` | SHA-256 (or the raw digest when prehashed) |
| `[1]` app root | `"{purpose}:{lowerhex(pubkey)}"` | 65B `r‖s‖recid` | keccak256 |
| `[2]` KMS root | `"dstack-kms-issued:" ‖ app_id(20B) ‖ app_root_pk(33B)` | 65B `r‖s‖recid` | keccak256 |

Link 3 is the one that matters, and it is why the KMS root key is a **caller-supplied parameter** rather than something the SDK fetches. Read it from the KMS being verified and an attacker who can answer that query can also mint a self-consistent chain. The docs point at `DstackKms.kmsInfo().k256Pubkey` or a pinned value, and say the same about `app_id`, since `AppInfo` comes from the CVM under test.

This is deliberately *not* wired into the simulator-backed test suites: the only way to obtain a KMS root key from the simulator is to recover it from the chain being checked, which is exactly the self-anchoring anti-pattern the API warns against. Chain coverage lives in the vector suite.

## The bug this would have shipped without shared vectors

`k256` **rejects** non-canonical high-S ECDSA signatures. Python's `cryptography` and Go's `decred/secp256k1` **accept** them.

A port done by eye would have silently reintroduced signature malleability in two of four languages — both `(r, s)` and `(r, n−s)` validating the same message, so a signature stops being a unique identifier. It is now an explicit check in all four, pinned by a dedicated negative vector.

This is not hypothetical for this repo: the Go compose-hash helper HTML-escaped `<` and hashed app-composes differently from every other SDK, and that digest is what gets whitelisted on chain. So `sdk/tests/vectors/signature_chain.json` is generated by `dstack/guest-agent/tests/signature_chain_vectors.rs` from the real KMS/guest-agent primitives, and all four SDKs assert against that one file. The generator re-derives and diffs on every run, so a format change fails there first.

## Dependencies

| SDK | Added |
|---|---|
| Python | **none** — `cryptography` + `eth-keys` already present |
| Go | **none** — stdlib `crypto/ed25519` + already-vendored decred secp256k1, reusing the existing keccak/recovery helpers |
| Rust | `k256`, `ed25519-dalek` (k256 already transitive via `alloy`) |
| JS | `@noble/curves`/`@noble/hashes` promoted optional-peer → real deps. `node:crypto` cannot verify a pre-hashed ECDSA digest, which `secp256k1_prehashed` requires. `bun.lock` realigned |

## Verification

`sdk/run-tests.sh` end to end against a freshly built simulator — **`EXIT=0`**:

- Rust 13 client + 9 tappd + 9 vector tests; both examples run
- Go `dstack` + `tappd`, including the three rewritten `TestSignAndVerify*`
- Python **147 passed**, `ruff` + `mypy` clean
- JS **113 passed** across 7 files

Against the rebuilt agent: `POST /Sign` still returns a 3-element chain; `POST /Verify` → `HTTP 404 {"error": "Service not found: Verify"}`.

`prek run --all-files` and `reuse lint` both pass. A proto/SDK sync audit found no leftover references and confirmed the other eight `DstackGuest` RPCs are still reachable in all four SDKs.

## Note for reviewers

`sdk/js/bun.lock` was hand-edited — only the workspace dependency categorization changed, resolved package entries are untouched. bun was not available in this environment, so a `bun install` to confirm would be welcome.


kvinwang:
Both points are fair. Pushed c8548a9 for the first.

**1. READMEs demonstrating the anti-pattern — fixed.**

You are right that this was the worst place for the inconsistency to land. The prose said "must be the app id you expect, not whatever `AppInfo` echoed back", and then every example did precisely that. The Go one was the clearest self-contradiction: `// pinned, or read from the DstackKms contract` on `kmsRoot`, with a bare `info.AppID` on the line directly above.

All four examples now make both anchors literals the caller owns:

```go
// Both anchors come from you, not from the CVM being checked.
appID, _ := hex.DecodeString("a9019d1b...")
kmsRoot, _ := hex.DecodeString("03...") // pinned, or read from the DstackKms contract
```

and each gained a short paragraph naming what the example deliberately does *not* do, since a comment inside a snippet travels with a copy-paste but a caveat three paragraphs up does not:

> Note what the example does *not* do: it never passes `info.AppID` from `client.Info()` straight through. That value is reported by the very CVM being verified, so a chain checked against it proves only that the CVM is self-consistent with itself. Use the app id you registered on chain, and if you want `Info` in the picture, compare it against that value rather than trusting it.

**2. recid 2/3 — no code change, but recorded.**

Agreed it is unreachable: it needs an ECDSA nonce whose `r` wrapped the curve order, around 2^-128, and no dstack signer has produced one. I did not want it to live only in this thread, though, so the vector generator now documents it under an explicit "what these vectors deliberately do NOT pin" heading — including *why* there is no fixture (it cannot be generated) rather than leaving a future reader to assume it was an oversight.

I considered narrowing Rust/Go/JS to `{0, 1}` for uniformity and decided against it: that would make three SDKs stricter than the format actually allows in order to match a limitation of one library, which is the wrong direction to resolve a divergence.

kvinwang:
Rebased onto `next` (21 new commits upstream). Force-pushed `ca5711d`.

**Conflicts: two, both from the same root cause.** #1107's `GetAttestationForAppKeyRequest` → `AttestAppKeyRequest` rename landed immediately adjacent to the `VerifyRequest`/`VerifyResponse` messages this branch deletes, so git could not tell which side to keep:

| File | Conflict | Resolution |
|---|---|---|
| `agent_rpc.proto` | upstream renames a message vs. this branch deletes the neighbouring ones | took both — kept the rename, kept the deletion |
| `rpc_service.rs` | same collision inside one `use` block | took upstream's import list minus the two Verify types |

The failure mode worth guarding against here is silently reverting the rename while resolving. It didn't happen: `git diff origin/next -- agent_rpc.proto` shows `message AttestAppKeyRequest {` as an *unchanged context line*, with only the Verify removal as a delta. `rpc_service.rs` is +3/−93, all of it Verify. `CHANGELOG.md` is +5/−0, so no upstream entry was displaced. No conflict markers survive anywhere in the tree.

**Re-verified against a simulator rebuilt from the rebased tree** — `sdk/run-tests.sh` `EXIT=0`:

- Rust 13 client + 9 tappd + 9 vector tests, both examples
- Go `dstack` + `tappd`
- Python **147 passed**
- JS **113 passed** across 7 files
- `prek run --all-files` and `reuse lint` clean

Live endpoint checks on the rebased agent:

```
POST /Verify                 → 404 {"error":"Service not found: Verify"}
POST /Sign                   → 200, 3-element signature_chain
POST /prpc/AttestAppKey      → 200   (upstream's rename intact)
```

That third line is the one I actually cared about. It confirms #1107's rework still works through my resolution rather than having been quietly clobbered — a green build alone would not have told us that, since the old name is gone from the proto either way.

PR #1112, opened 2026-08-24
feat(guest-agent): collect on-demand GPU attestation evidence

kvinwang:
## Problem

`GpuInfo` replays a record written at boot and cannot answer whether a GPU can produce fresh attestation evidence now. A driver reload can leave a device that still responds to NVML but can no longer attest.

## Fix

Add `AttestGpu`, which collects fresh GPU evidence for a caller-chosen 32-byte nonce.

The response is vendor-neutral:

```proto
message AttestGpuResponse {
  repeated GpuEvidenceBundle bundles = 1;
}

message GpuEvidenceBundle {
  string vendor = 1;
  string format = 2;
  bytes evidence = 3;
}
```

The current NVIDIA implementation returns one bundle:

- `vendor`: `nvidia`
- `format`: `nvidia-nvattest-collect-evidence-json-v1`
- `evidence`: the opaque `nvattest collect-evidence` evidences array

The agent only collects evidence. It does not run the local NVIDIA verifier or return an appraisal. A relying party selects a verifier using `vendor` and `format`, then verifies the signature, certificate chain, measurements, and nonce embedded in the evidence.

The response intentionally does not echo the nonce because the trusted nonce binding is inside the GPU-signed evidence. It also does not claim that the GPU is attached to this TD: current GPU evidence remains relayable until TDISP/TEE-IO provides device binding.

Evidence collection is serialized to avoid concurrent access to the same devices, but it is not rate-limited because `collect-evidence` does not fetch OCSP or RIM collateral from NVIDIA.

The API is exposed by the Rust, Python, Go, JavaScript, and curl SDK documentation.

## Shared crate

The boot gate previously owned the NVIDIA CLI invocation, proxy validation, timeout, and nonce checking. This PR extracts that code into the `nvattest` crate and rewires `dstack-util` to use it. `run_command` uses `kill_on_drop`, so a timed-out or cancelled tool does not outlive its caller.

## Verification

- `cargo check -p nvattest -p dstack-guest-agent`
- `cargo test -p nvattest`: 9 passed
- `cargo test -p dstack-guest-agent gpu_attest`: 3 passed
- `cargo check` in `sdk/rust`
- `go test ./dstack -run TestAttestGpu`
- `cargo fmt --all --check`
- `git diff --check`
- Python source compilation passed; pytest collection is blocked by the unavailable `evidence_api` dependency.
- JavaScript tests were not run because dependencies are not installed in the review environment.


kvinwang:
Adding a correction to the framing above, from checking the actual artifact rather than reasoning from the claim names.

The PR body justifies "not remote evidence" via the relay argument. There is a blunter reason that comes first: **the output is unsigned.**

`nvattest --verifier local` returns the verifier's *conclusion*, not the GPU's signed report. From this repo's own captured H100 output (`dstack-util/tests/fixtures/gpu_attestation_h100.json`):

- the embedded `detached_eat` is `{"alg":"none"}` with an **empty signature**, `iss: NVAT-LOCAL-VERIFIER`
- `x-nvidia-gpu-attestation-report-signature-verified: true` is an assertion about a check already performed, not proof anyone can redo
- `x-nvidia-gpu-attestation-report-cert-chain` holds verdicts (`x-nvidia-cert-ocsp-status: "good"`), not the chain

The GPU's signed SPDM report and certificate chain are consumed during verification and never appear in the output. So a third party handed this JSON has nothing to check at all — relay or no relay.

This does not change the design, but it does sharpen why it is drawn this way, and it explains something worth stating explicitly: the boot-time path is sound with *the same unsigned artifact*, because its trust does not come from the JSON self-authenticating. It comes from measured dstack code (pinned by `os_image_hash`) having appraised the GPU before any workload existed, with sha256 of the exact bytes in RTMR3 under the quote. **The appraiser is what is trusted, not the artifact.**

`11c1e40f2b` states both reasons in the proto, all four SDKs, the curl reference and the CHANGELOG, and adds a test pinning the format claim against the fixture — if a future SDK starts signing the EAT, that test fails instead of the API docs quietly becoming wrong.

If you want third-party-verifiable GPU evidence later, there are two routes, both larger than this PR: `nvattest collect-evidence` emits the raw signed report and cert chain for someone else to appraise, and `--verifier remote` returns an NVIDIA-signed EAT from NRAS. Neither gives TD binding — that still needs TDISP/TEE-IO — but either would at least let a remote party check the GPU's own signature.

PR #1113, opened 2026-08-24
guest-agent: freeze the unversioned API at the v0.5.11 surface

kvinwang:
## Problem

The `next` branch made breaking changes on the unversioned guest-agent paths that 0.5.x SDKs call: the `Verify` RPC was deleted (all four 0.5.11 SDKs expose `verify()`), and `EmitEvent` was deleted such that a pre-0.6 client gets a generic unknown-method HTTP 400 — prpc answers "no such method" and "handler failed" identically, so the caller cannot tell a deliberate removal from talking to the wrong socket.

With the versioned `/v1` API arriving in 0.6.0 (follow-up PR), the unversioned surface's job changes: it must stay wire-compatible with v0.5.x clients indefinitely, and all new capability moves to v1.

## Fix

- Restore `Verify` with its exact v0.5.11 wire shape and semantics (ed25519 / secp256k1 / secp256k1_prehashed; k256's canonical low-S-only parsing preserved). Doc-commented as legacy-only: signature verification needs no TEE, and the RPC exists solely so 0.5.x clients keep working. It will not appear in v1.
- Restore `EmitEvent` as a deliberate always-failing stub: runtime RTMR3 events are system-owned in 0.6, and the method now fails with a self-explanatory message instead of an indistinguishable 400.
- Annotate the service as frozen: no renumbering, no removals, no semantic changes; new functionality goes to `dstack.guest.v1`.

Sanctioned behavior-only changes are kept: `GetQuote` fails closed on non-TDX platforms, `GetTlsKey` validates `not_before < not_after`.

## Verification

- Method-set diff against `git show v0.5.11:guest-agent/rpc/proto/agent_rpc.proto`: zero v0.5.11 `DstackGuest` methods missing.
- 6 new tests: sign→verify round trips for all three algorithms through the real handlers, tampered-data negative case, unsupported algorithm, and `emit_event_reports_its_removal`. `cargo test -p dstack-guest-agent --all-features`: 84 passed.
- `cargo fmt --all --check` clean, `cargo clippy -- -D warnings --allow unused_variables` clean, `cargo check --workspace --all-features` clean.


kvinwang:
Folded into #1114 — the freeze commits are part of that branch's history; reviewing the 0.6.0 API change as one PR.

Issue #1094, opened 2026-08-21
0.6 prerelease checklist: stabilize guest-agent API and SDKs


h4x3rotab:
## Objective

Ship dstack 0.6 with a stable, versioned guest-agent API and consistent Go, JavaScript/TypeScript, Python, and Rust SDKs. Existing 0.5.x clients must continue to work against a 0.6 agent.

## Compatibility

- [x] Keep the unversioned API fully compatible with 0.5.x, verified with v0.5.10 generated clients and SDKs. — #1121 runs the released SDK suites against the current agent in CI. The pinned tags are **v0.5.8 and v0.5.11**, not v0.5.10: `v0.5.10:sdk` and `v0.5.11:sdk` are the same tree object, so running both bought nothing. Note the limit — old JSON clients ignore unknown keys, so this job catches *breaking* changes to the frozen surface but not *additive* ones; the shape freeze in `frozen_surface` is what covers additions.
- [x] Keep `GetQuote` unchanged: the v0.5 response fields and semantics remain the complete contract. Cross-platform attestation is exposed through `Attest`, not `GetQuote`.
- [x] Keep `GetKey` wire- and behavior-compatible. Publish a normative specification for path, purpose, supported algorithms, defaults, output encoding, and signature-chain verification. — spec in #1123 (`docs/guest-api-v0.md`), byte-level: HKDF salt, what enters the KDF vs only the claim, both chain-link preimages, all three signing modes, and the hazard that `algorithm` does not domain-separate.
- [x] Make `EmitEvent` return 404 with informative error message (deprecated method)
- [x] Unknown algorithms must be rejected.

## Attestation

- [x] Make `Attest` the single API for platform and GPU attestation

## SDKs

- [ ] Release all changed SDKs as 0.6.0 and apply semantic versioning consistently.
- [x] Expose the same `TlsKeyOptions` fields and defaults in every SDK. `usage_server_auth` defaults to `true`; remove the unused `path` option. — #1120
[...]

PR #1116, opened 2026-08-24
guest-agent: freeze the v0 API at v0.5.11 and add the versioned v1 API

kvinwang:
Single PR for the dstack 0.6.0 guest-agent API: freezes the unversioned surface at exactly v0.5.11, introduces the versioned v1 API, and switches the four SDKs to mirror both surfaces. Supersedes #1113/#1114.

## Problem

`next` had broken the unversioned paths that 0.5.x clients call (`Verify`/`EmitEvent` deleted, `Worker.GetAttestationForAppKey` renamed), while the unversioned API carried design debt it could not shed without further breakage: `GetTlsKey` names a key what is really certificate issuance; `GetKey`'s `purpose`/`algorithm` are accepted but do not domain-separate the derived key (one 32-byte secret serves both curves); `Sign`/`Verify` are pure computation granting no capability `GetKey` does not already grant; `Info` returns triple-nested JSON duplicating its own top-level fields; and the chain claim `keccak256("{purpose}:{hex(pubkey)}")` lets a caller-chosen `purpose` mint nearly arbitrary signed strings.

## Fix

Every surface becomes a frozen v0 fossil plus a v1, selected by URL path only (no header negotiation):

| Listener | v0 (frozen = v0.5.11) | v1 |
|---|---|---|
| dstack.sock | `/` (alias), `/v0`: 9 v0.5.11 methods | `/v1`: IssueCert, GetKey, Attest, AttestGpu, Info, Version |
| external | `/prpc` (alias), `/prpc/v0`: Info, Version, GetAttestationForAppKey | `/prpc/v1`: Info, Version, Health |
| tappd.sock | Tappd unversioned (predates v0) | — |

**v0**: `Verify` restored byte-identical to v0.5.11; `EmitEvent` restored as a deliberate always-failing stub (RTMR3 runtime events are system-owned; the message is the signal, since prpc reports handler failure and unknown method differently only in the body); `GetAttestationForAppKey` restored TDX-only. Never-released post-0.5.11 additions removed. A descriptor-digest test pins the frozen shape so even a wire-compatible field addition fails CI.

**v1** keeps only what needs the TEE — derivation, attestation, and KMS-backed cert issuance:
- `GetKey(domain, algorithm)`: `HKDF-SHA256(salt="dstack-guest-v1", IKM=app_root_k256, info=LP("dstack-guest-v1-key")||LP(algorithm)||LP(domain))` — length-prefixed encoding, per-algorithm keys, v1-specific salt so no v0 input can reproduce a v1 key. Chain claim binds `(algorithm, domain, raw pubkey)` and is structurally unforgeable from the v0 claim format (a v0 preimage's tail is lowercase hex; the v1 length prefix contains 0x00). Constants and vectors live in `ra_tls::api_v1`, shared by agent and tests.
- `IssueCert`: honest name for what `GetTlsKey` did — certificate issuance via KMS `SignCert`, generating a fresh key as the integrated one-step mode.
- `Attest` (sole CVM attestation entry; optional boot-time GPU evidence), `AttestGpu` (on-demand, 32-byte nonce), flat identity-and-configuration `Info` (attestation data deliberately excluded — it belongs to `Attest`, quote-backed), `Version`.
- No Sign/Verify/EmitEvent/GpuInfo/AttestAppKey: signing and verification are the app's and relying party's job per the spec; there is no v1 key the external AttestAppKey could honestly attest.
- Gateway health polling moves to `/prpc/v1/Health`. Identity for `Info` is decoded once at startup (an anonymous external caller can no longer drive hardware quote generation per request).

**SDKs** (all four): `ClientV0` mirrors the frozen surface exactly (v0.5.11 method set, `verify()` back, `emit_event()` surfaces the removal error); `ClientV1` mirrors the six v1 methods; the never-released local `verify_signature`/`verify_signature_chain` helpers and their vectors are removed — the normative spec is the verification reference. Deprecated aliases keep existing code compiling. Versions unified at 0.6.0.

**Spec**: `docs/guest-api-v1.md` — byte-level KDF/claim encodings, verification steps, status-code/probe rules (measured: absent mount → 404 HTML, unknown method → 404 JSON, failed handler → 400 JSON), migration contract including that v1 keys deliberately differ from v0 keys for the same inputs.

## Verification

- Acceptance diff vs the `v0.5.11` tag (comments stripped): frozen `DstackGuest`, `Worker`, `Tappd` identical; 16/19 messages byte-identical, 3 differ by `reserved` only.
- `./dstack/run-tests.sh` (agent-backed simulator): **1235 passed, 0 failed**. `sdk/run-tests.sh`: all four suites green (Rust 38, Go 51, Python 156, JS 131).
- Committed KDF/claim vectors; `a_v0_path_cannot_reproduce_a_v1_key`; forgery test asserts the structural hex/0x00 property; frozen-descriptor test verified to catch a planted field addition; route tests prove `/v0` and unversioned are the same handler and each mount serves only its own surface.
- fmt/clippy (per CLAUDE.md)/workspace check clean; Go vet/gofmt clean; JS tsc/tsup clean.
- An 8-angle adversarial self-review ran against this branch; all 26 confirmed findings were fixed here or explicitly dispositioned.

## Known follow-ups (out of scope)

- Remaining #1094 SDK cleanups (`asUint8Array`→`toPkcs8Der`, Go `usage_server_auth` default, JS transport status checks).
- CI job running pinned v0.5.10/v0.5.11 released SDKs against the new agent (compat regression).
- Pre-existing simulator fixture inconsistency: `sdk/simulator/attestation.bin` was captured from a different app-compose than `app-compose.json` (`compose_hash` mismatch, reproducible on v0 too) — needs a fixture refresh.


docs/app-health-checks.md
```diff
@@ -46,7 +46,7 @@ older guest chokes on.
 ## Where the verdict comes from
 
 The guest agent recomputes a verdict every 5 seconds and caches it; the gateway
-polls `Worker.Health`, which only reads that cache. The two cadences are
+polls `WorkerV1.Health` (at `/prpc/v1`), which only reads that cache. The two cadences are
```
kvinwang:
```suggestion
polls `/prpc/v1/Health`, which only reads that cache. The two cadences are
```

docs/guest-api-v1.md
```diff
@@ -0,0 +1,668 @@
...
+returns, and verify locally following this document. A v0 client may keep calling
+the unversioned `Verify` for single-signature checks; the v0 `Sign` RPC's
+per-algorithm signing modes are documented on that surface, not here.
+
+`GetQuote` is absent because `Attest` subsumes it. `GetQuote` answers on Intel TDX
+and nowhere else, and the `VersionedAttestation` that `Attest` returns already
+carries the TDX quote and the event log. See
+[Extracting a quote from an attestation](#extracting-a-quote-from-an-attestation).
+
+`GpuInfo` is absent from v1, and has also been removed from the unversioned
```
kvinwang: GpuInfo加了又删，没必要在文档留痕迹

dstack/guest-agent/rpc/proto/agent_rpc_v1.proto
```diff
@@ -0,0 +1,445 @@
...
+// unattested, so a relying party gains nothing over checking it itself.
+// Applications sign locally with a standard library, using the key `GetKey`
+// returns; `docs/guest-api-v1.md` specifies chain verification normatively.
+//
+// There is no `EmitEvent` either: runtime RTMR3 events are system-owned as of
+// 0.6.0 and cannot be extended by an app.
+//
+// The unversioned surface keeps `Sign` and `Verify` for v0.5.x clients. They
+// are frozen there and are not carried forward.
+service DstackGuestV1 {
```
kvinwang: 既然已经在namespace里了，是不是就叫DstackGuest好些？

dstack/guest-agent/rpc/proto/agent_rpc_v1.proto
```diff
@@ -0,0 +1,445 @@
...
+//
+// There is deliberately no `AttestAppKey`. v0's version attests the key its own
+// KDF derives at path `vms` with purpose `signing`, and no v1 `GetKey(domain,
+// algorithm)` can return that key -- so a pure-v1 application could be handed
+// an attestation of a public key whose private half it has no way to obtain.
+// A v1 application attests its own key instead: derive it at `/v1/GetKey`,
+// commit the public key into `report_data`, and call `/v1/Attest`. The
+// application then hands that attestation to its relying parties. Legacy flows
+// keep using the frozen `Worker.GetAttestationForAppKey`.
+service WorkerV1 {
```
kvinwang: 就叫Worker?

dstack/guest-agent/rpc/src/generated.rs
```diff
@@ -4,3 +4,14 @@ pub const FILE_DESCRIPTOR_SET: &[u8] =
     include_bytes!(concat!(env!("OUT_DIR"), "/file_descriptor_set.bin"));
 
 include!(concat!(env!("OUT_DIR"), "/dstack_guest.rs"));
+
```
kvinwang: 旧版也建个mod v0包起来？

dstack/kms/src/crypto.rs
```diff
@@ -4,8 +4,8 @@
 
 use anyhow::{Context, Result};
 use k256::ecdsa::SigningKey;
-use sha3::{Digest, Keccak256};
 
+use ra_tls::guest_api_v1::sign_recoverable_keccak256;
```
kvinwang: 最好改名为 `ra_tls::api_v1::sign_recoverable_keccak256`。ra_tls不应该有guest api知识。

dstack/kms/src/crypto.rs
```diff
@@ -35,11 +35,9 @@ pub(crate) fn sign_message(
...
     message: &[u8],
 ) -> Result<Vec<u8>> {
-    let digest = Keccak256::new_with_prefix([prefix, b":", appid, message].concat());
-    let (signature, recid) = key.sign_digest_recoverable(digest)?;
-    let mut signature_bytes = signature.to_vec();
-    signature_bytes.push(recid.to_byte());
-    Ok(signature_bytes)
+    // Same 65-byte `r || s || v` envelope every dstack chain link uses; the
+    // preimage is what differs between them.
+    sign_recoverable_keccak256(key, &[prefix, b":", appid, message].concat())
```
kvinwang: 这里换了实现后会不会破坏兼容？比如kms给已经部署的CVM发的key发生变化？
最好在PR 841中增加这类用例。

dstack/kms/src/crypto.rs
```diff
@@ -52,12 +50,10 @@ pub(crate) fn sign_message_with_timestamp(
     message: &[u8],
 ) -> Result<Vec<u8>> {
     let timestamp_bytes = timestamp.to_be_bytes();
-    let digest =
-        Keccak256::new_with_prefix([prefix, b":", appid, &timestamp_bytes[..], message].concat());
-    let (signature, recid) = key.sign_digest_recoverable(digest)?;
-    let mut signature_bytes = signature.to_vec();
-    signature_bytes.push(recid.to_byte());
-    Ok(signature_bytes)
+    sign_recoverable_keccak256(
```
kvinwang: 同上

PR #1134, opened 2026-08-25
docs: add a release-candidate testing runbook

kvinwang:
## Problem

Testing 0.6.0-rc0 on real TDX and SEV-SNP hardware cost more time in detours than in testing. Most of them were not hard problems — they were things that are obvious once seen and invisible beforehand:

- `/dev/tdx_guest` does not exist on a TDX *host*, so checking for it reports a perfectly good host as having no TDX. I concluded a machine could not run the test, and scoped around a limitation that was not there.
- On the SEV-SNP host, `vhost_vsock` was not loaded *and* VMware's vsock transport was claiming the stack, so the VMM could not bind its host API. Neither is a dstack problem, and both look like one.
- `/dev/sev` is root-only by default; after fixing group membership, a detached supervisor still held the old groups, so QEMU kept failing after the fix appeared to work.
- `df -h /` said 12 GB free while `/home` was a 7 TB array, which produced a wrong "we are too tight on disk" plan.

None of this is written down anywhere, and the next person on different hardware starts from zero.

## What this adds

`docs/rc-testing-runbook.md` — the order to do things in, and the specific traps, with the commands that actually work.

Beyond host setup it covers the parts that are easy to get wrong or dangerous to get wrong:

- **`~/.dstack-vmm/config.json` defaults `vmm-cli.py` at a production VMM.** On both machines used last round it pointed at a live deployment with credentials. Any `deploy` omitting `--url` lands there. This one is worth reading before touching anything.
- Which tags publish to Docker Hub, so a private-registry test does not accidentally publish.
- That the container Dockerfiles clone from GitHub rather than the working tree, so the commit must be pushed before images are built.
- That `release/**` is protected against force-push and deletion, so a rebase-based workflow will be rejected.
- ACME config lives in the KvStore behind the admin API in 0.6.0; the `CF_API_TOKEN` and `ACME_STAGING` variables still present in `gateway/dstack-app/deploy-to-vmm.sh` are vestigial and do nothing.
- The four places the version number lives, only two of which are cross-checked.

## Two sections worth reading even if the rest is skimmed

**What each verification actually proves.** A run where every quote is accepted cannot distinguish "verification passed" from "verification never ran". Last round supplied the negative case by accident — the AMD CVMs produced 7 rejections alongside 6 grants in the same KMS log — but if everything passes, the rejection needs to be arranged deliberately. The table says what each check buys.

**How to read DNS evidence without fooling yourself.** `dig +short` prints nothing for both NODATA and timeout, which are usually the two hypotheses being separated; certbot deletes the challenge record after issuing, so its later absence is cleanup rather than a fault; and Cloudflare needs more than ten seconds to serve a new record from its own authoritative edge. Each of these produced a wrong reading during the last round before being caught. The section also covers running a control through the same path as the test — four public proxies were tried for an independent view of AMD KDS and every one returned the same error for `github.com` as for the target, which without the control reads as evidence about AMD.

## AMD SEV-SNP

Section 8 recommends provisioning host certificates as part of setup rather than relying on AMD KDS. `kdsintf.amd.com` is a single global endpoint with no mirror, is aggressively rate-limited, and was entirely unreachable during this round — TCP timeouts and 100% ICMP loss from five vantage points across four autonomous systems and two continents, while everything else answered normally and DNS stayed healthy. dstack already reads ASK and VCEK from the SNP extended report's certificate table, and AMD's own specification recommends host-side caching, so a provisioned host removes the dependency entirely.

## Verification

`prek run --files docs/rc-testing-runbook.md` passes. `docs/**/*.md` is already covered by `REUSE.toml`, so no licensing registration was needed.

Every command in the document was run during the 0.6.0-rc0 round; the host-capability checks in section 1 were re-run against both hosts while writing it rather than transcribed from memory.

### DIFF
--- a/docs/rc-testing-runbook.md
+++ b/docs/rc-testing-runbook.md
@@ -1,0 +1,313 @@
+# Release-candidate testing runbook
+
+What to do, in order, to exercise a dstack release candidate on real TEE
+hardware — and the specific things that cost time the last round, so they cost
+nothing the next one.
+
+Written after testing 0.6.0-rc0 on an Intel TDX host and an AMD SEV-SNP host.
+Assumes fresh machines: nothing here depends on state left behind by that run.
+
+---
+
+## 1. Confirm the host can actually do what you think
+
+Get this wrong and you will scope the whole test around a capability you have,
+or skip one you don't.
+
+### Intel TDX host
+
+```bash
+cat /sys/module/kvm_intel/parameters/tdx      # expect: Y
+grep -o -m1 tdx_host_platform /proc/cpuinfo   # expect: tdx_host_platform
+systemctl is-active qgsd pccs                 # expect: active active
+```
+
+**`/dev/tdx_guest` does not exist on a TDX host and its absence means nothing.**
+That device is the *guest*-side attestation interface; it appears inside a TD,
+not on the machine running them. Checking for it will tell you a perfectly good
+TDX host has no TDX. Use the KVM module parameter and the CPU flag.
+
+`qgsd` is what turns a TD's attestation request into a signed quote, and `pccs`
+caches the Intel collateral used to verify it. Without both, CVMs boot but
+cannot obtain keys.
+
+### AMD SEV-SNP host
+
+```bash
+cat /sys/module/kvm_amd/parameters/sev_snp    # expect: Y
+grep -o -m1 sev_snp /proc/cpuinfo             # expect: sev_snp
+ls -l /dev/sev                                # must exist AND be accessible
+qemu-system-x86_64 -object help | grep sev-snp-guest
+lsmod | grep -E '^vhost_vsock|vmw_vsock_vmci'
+```
+
+Three host-level problems each blocked us for a while:
+
+- **`vhost_vsock` not loaded.** The VMM binds its host API on vsock CID 2 and
+  fails with `Cannot assign requested address (os error 99)`. Fix: `sudo modprobe
+  vhost_vsock`. Does not survive reboot — add `/etc/modules-load.d/` to persist.
+
+- **VMware's vsock transport hijacking the stack.** If `vmw_vsock_vmci_transport`
+  is loaded (it often is, by default), it claims the vsock transport and CID 2
+  binding still fails even with `vhost_vsock` present. Fix: `sudo modprobe -r
+  vmw_vsock_vmci_transport vmw_vmci`. It comes back on reboot — blacklist to
+  persist. This one is easy to misdiagnose as a dstack bug.
+
+- **`/dev/sev` is `root:root 0600` out of the box** and your user is probably not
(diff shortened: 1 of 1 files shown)


kvinwang:
Added a gateway-clustering section, and corrected a claim the first version made about proxy coverage.

## Clustering (new section 6)

Testing a two-node cluster after opening this turned up a trap worth writing down.

**Sync is already on in a single-node run.** `entrypoint.sh` sets
`SYNC_ENABLED=$([ "$NODE_ID" -gt 0 ] && ...)`, so a node with `NODE_ID=1` logs
`WaveKV: detected certificate changes` and looks clustered while nothing has ever
replicated. I had assumed the opposite — that sync was off — and wrote as much
before checking the entrypoint. Neither reading is safe without looking.

**Every node must deploy under the same `--name`.** Cluster members require the
peer's `app_id` to match their own (`kv/https_client.rs`), which is deliberate:
it confines WaveKV replication to instances of one authorized compose. `app_id`
is the compose hash and `name` is part of it, so a second node deployed under a
different name fails with:

```
app_id mismatch: expected c90cc75a6ceb…, got aa04b7bd7875…
```

That reads like a trust or networking fault and is neither. Per-node values go in
`--env-file`, whose values are encrypted at deploy time and excluded from the
hash — only key names are recorded, as `allowed_envs`. Diffing the two compose
files showed `name` as the *only* differing field, which is what made the cause
obvious.

The section also lists what actually proves a cluster works. Agreeing peer lists
are the weakest signal; the one that matters is traffic entering node2 and
reaching an app whose WireGuard tunnel terminates on node1, which shows the
replicated state is usable routing rather than metadata that happens to match.

## Correction to section 7

The first version listed "`curl` through the proxy to the app — end to end,
including TLS termination and routing". True, but it invites the reading that the
proxy is covered. It is not: ingress maps as `<id>[-[<port>][s|g]].<domain>` and
the suffixes select different code paths — `s` is TLS passthrough through
`tls_passthough.rs` with its own `_dstack-app-address` TXT resolution, `g` is
HTTP/2. The rc0 round exercised only the no-suffix path, and only with a small
request. The table now says so, and a subsection spells out what remains untested.

Also noted there: a passthrough-resolution error while testing the *terminated*
path usually means the gateway had no certificate and fell through to SNI
routing. I spent time treating that as a DNS problem before realising the missing
certificate was upstream of it.

## Housekeeping

Added the `pgrep -f` self-match trap, which cost the most time of anything in the
run despite being the least interesting. A waiter looping on
`pgrep -f "build-image.sh ..."` never exits while you check progress with a
command whose own command line contains that string — **checking kept it
waiting**, and an image that had been ready for half an hour sat unpushed. The
same self-match reported a finished build as running and, earlier, made a bare
`pgrep -af qemu-system` match the grep itself and report a VM that did not exist.
Check for the artifact, not the process.

PR #1145, opened 2026-08-26
feat(vmm): run virtio-net on vhost-net with configurable queue pairs

kvinwang:
## Problem

A CVM's virtio-net data plane was limited by QEMU's single main-loop thread. At higher packet rates the host TAP dropped traffic before it reached the guest, so guest counters stayed clean while throughput hit one core's ceiling.

Also, bridge interfaces had two owners. Depending on filtering and queue count, either `qemu-bridge-helper` or netd created the TAP. That made vhost capability, multiqueue preparation, teardown, and recovery depend on which path a node happened to take.

## Data plane

The node and each VM can select the vhost-net data plane, and each VM can select a queue-pair count:

```toml
[cvm]
max_net_queues = 16

[cvm.networking]
vhost = true
```

```bash
vmm-cli.py deploy ... --net bridge --net-queues 4
vmm-cli.py deploy ... --net bridge --net-no-vhost
```

Vhost is off in the shipped configuration for upgrade compatibility. With vhost enabled, queues default to the VM's vCPU count capped at 16 and `cvm.max_net_queues`; without vhost they default to one. An explicit queue count is preserved with either vhost setting. The hard queue limit is 64.

For `N > 1`, QEMU receives matching backend and device settings:

```text
-netdev tap,...,vhost=on,queues=N
-device virtio-net-pci,...,mq=on,vectors=2N+2
```

Macvtap opens `/dev/tapN` once per queue pair and passes `fds=a:b:...`. User networking has no vhost or multiqueue backend. Custom mode owns its complete netdev string and reports no inferred data-plane state.

Neither vhost nor queue count is measured, so retuning a NIC does not change application identity.

## One owner for host interfaces

Every bridge and macvtap NIC is now prepared by netd, including unfiltered single-queue bridge NICs. `qemu-bridge-helper` is no longer used, and `/etc/qemu/bridge.conf` no longer needs an `allow` entry.

This is an operational compatibility change: **netd must be upgraded and running before the VMM on every host that runs bridge or macvtap VMs.** User and custom networking do not require it.

The VMM sends netd the required queue count and verifies that netd reports the same count. A mismatch, including an old netd that omits the field, rolls back preparation and fails the launch with a clear error.

## Interface teardown

Stop, network update, and removal ask netd to remove each bridge or macvtap NIC recorded in the VM's runtime network snapshot. `remove` no longer carries a `filtered` flag: netd clears any nwfilter binding at that name best-effort. One-shot `dstack-vmm run` does not manage netd interfaces, so it refuses bridge and macvtap NICs outside `--dry-run`.

## API and UI

- `NetworkingConfig`: optional `vhost` and `queues`
- `NetworkInterfaceStatus`: effective `vhost`, `queues`, and macvtap mode
- `NetworkingCapabilities`: node `default_vhost` and `max_queues`
- Deploy/update CLI and web UI expose per-NIC vhost and queue selection

Running VMs report the data plane captured at launch. Stopped VMs report what the next launch would resolve from their manifest and current node configuration. A running VM started by an older VMM with no runtime snapshot is conservatively reported as vhost off with one queue.

## Split

Two follow-up PRs are stacked on this branch:

- #1213: port mappings can name the NIC they enter through (`PortMapping.nic_index`, `--port ...@<nic>`)
- #1214: netd interface lifecycle rework (ownership records, whole-VM sweeps, `netd list`/`remove-vm`/`remove-interface`, removal retry, per-VM launch lock)

## Verification

- `cargo test -p dstack-vmm`: 175 passed
- `cargo clippy -p dstack-vmm --tests`: no warnings
- Real TDX deployments (on the pre-split branch) covered vhost on/off, bridge multiqueue, macvtap multiqueue, libvirt filtering, DHCP/SSH, queue negotiation, and throughput. See `docs/network-data-plane.md` for the measurements and operational guidance.

### DIFF
--- a/docs/bridge-networking.md
+++ b/docs/bridge-networking.md
@@ -4,7 +4,7 @@ By default, dstack-vmm uses **user** networking (QEMU's built-in SLIRP stack, no
 
 ## When to use bridge networking
 
-- High connection concurrency (passt becomes CPU-bound at ~25K+ concurrent connections)
+- High connection concurrency (user-mode networking becomes CPU-bound at ~25K+ concurrent connections)
 - Workloads that need full L2 network access
 - Environments where VMs need to be directly reachable on the LAN
 
@@ -21,11 +21,11 @@ bridge = "virbr0"
 ### Per-VM override
 
 Individual VMs can override the global networking mode via:
-- **CLI**: `vmm-cli.py deploy --net bridge` or `--net passt`
+- **CLI**: `vmm-cli.py deploy --net bridge`, `--net user`, or `--net macvtap`
 - **Web UI**: Networking dropdown in the deploy dialog
 - **API**: `networking: { mode: "bridge" }` in `VmConfiguration`
 
-Only the mode is per-VM; the bridge interface name always comes from the global config.
+The bridge interface name comes from the global config unless the node lists it in `cvm.allowed_bridges`. VMs may also override the vhost and queue settings — see [network-data-plane.md](network-data-plane.md).
 
 ## Host setup
 
@@ -143,9 +143,15 @@ mode = "bridge"
 bridge = "dstack-br0"
 ```
 
-### QEMU bridge helper setup (required for both options)
+### QEMU bridge helper setup (needed unless every bridge NIC goes through netd)
 
 The bridge helper allows QEMU to create and attach TAP devices without VMM needing root privileges.
+It is used only on the single-queue bridge paths; a NIC that `netd` builds never touches it, so a
+node that runs `netd` for all of its bridge VMs does not need it at all.
+
+The VMM probes `/usr/lib/qemu/qemu-bridge-helper`, `/usr/libexec/qemu-bridge-helper` and
+`/usr/local/libexec/qemu-bridge-helper`. Set `cvm.qemu_bridge_helper` in `vmm.toml` for a path
+outside that list.
 
 ```bash
 # Allow QEMU to use the bridge
@@ -159,12 +165,13 @@ sudo chmod u+s /usr/lib/qemu/qemu-bridge-helper
 
 ## How it works
 
-- VMM passes `-netdev bridge,id=net0,br=<bridge>` to QEMU
-- QEMU's bridge helper (setuid) creates a TAP device and attaches it to the bridge
+- With more than one queue pair, or with libvirt filtering on, `netd` creates the TAP and the VMM passes `-netdev tap,id=net0,ifname=<tap>,...` — this is the usual case on a node running `netd` with multi-vCPU VMs, since queue pairs default to the VM's vCPU count. Without `netd`, a bridge NIC that took that default drops back to one queue pair and takes a helper path below
+- Otherwise the VMM passes `-netdev tap,id=net0,br=<bridge>,helper=<qemu-bridge-helper>,vhost=on`, or `-netdev bridge,id=net0,br=<bridge>` when vhost is off or no helper is found
+- QEMU's bridge helper (setuid) creates a TAP device and attaches it to the bridge on the two helper paths
 - Guest MAC address is derived from SHA256 of the VM ID, with an optional configurable prefix (stable across restarts for DHCP IP consistency)
 - The host DHCP server (dnsmasq) assigns an IP to the VM
-- When QEMU exits, the TAP device is automatically destroyed
-- VMM does not need root or `CAP_NET_ADMIN`
+- On the two bridge-helper paths the TAP disappears when QEMU exits; a `netd`-created TAP is persistent and is deleted when the VMM tears the VM's networking down
+- The VMM process itself needs neither root nor `CAP_NET_ADMIN` on any path; the `netd` path moves that privilege into a separate root service instead
 
 ### MAC address prefix
 
@@ -194,13 +201,15 @@ The remaining bytes are derived from the VM ID hash. The prefix applies to all n
 
 ### Mixing networking modes
 
-Bridge and passt VMs can coexist. Set the global default in `vmm.toml` and override per-VM as needed:
+Bridge and user-mode VMs can coexist. Set the global default in `vmm.toml` and override per-VM as needed:
 
 ```bash
-# Global default is bridge, but deploy this VM with passt
-vmm-cli.py deploy --name my-vm --image dstack-0.5.6 --compose app.yaml --net passt
+# Global default is bridge, but deploy this VM with user networking
+vmm-cli.py deploy --name my-vm --image dstack-0.5.6 --compose app.yaml --net user
 ```
 
-### vhost-net and TDX
+### vhost-net and multiqueue
 
-vhost-net (kernel data plane offload for virtio-net) is **not enabled** for bridge mode. TDX encrypts guest memory, which prevents the host kernel from performing DMA-based packet offload. The default QEMU userspace virtio backend is used instead.
+Bridge NICs can run on the host kernel's vhost-net data plane and expose several virtio-net queue pairs. Both are off by default and enabled per node or per VM — see [network-data-plane.md](network-data-plane.md) for the knobs, the enablement checklist, the mode support matrix, and how to pick a queue count.
+
+vhost-net works in a TDX guest: the virtio rings and buffers live in shared, unencrypted memory so that a host-side backend can reach them, which is the same mechanism `vhost-vsock-pci` has always relied on.
--- a/docs/libvirt-network-filter.md
+++ b/docs/libvirt-network-filter.md
@@ -14,8 +14,11 @@ host mechanism.
 
 The measurable acceptance criteria are:
 
-- `network_filter = "none"` preserves the existing QEMU `-netdev bridge`
-  behavior and does not require `netd` or libvirt.
+- `network_filter = "none"` installs no nwfilter binding. It still uses `netd`
+  for any NIC with more than one queue pair, and a `tap` netdev behind
+  `qemu-bridge-helper` whenever vhost is on; only a single-queue, non-vhost
+  bridge NIC keeps the historical `-netdev bridge` path with no `netd` or
+  libvirt dependency.
 - `network_filter = "libvirt"` creates the TAP and filter binding before QEMU
   is submitted to Supervisor, and uses QEMU `-netdev tap`.
 - A failed TAP or filter setup prevents QEMU from starting and rolls back all
@@ -53,8 +56,11 @@ allowed_macvtap_parents = []
 
 Macvtap is excluded from `allowed_network_modes` by default. Empty bridge and
 macvtap-parent allowlists prevent RPC callers from overriding the respective
-node defaults. If macvtap is explicitly enabled, callers may select only a
-parent in `allowed_macvtap_parents`; the macvtap forwarding mode always comes
+node defaults. If macvtap is explicitly enabled, callers may select a
+parent listed in `allowed_macvtap_parents`, the node's own configured parent, or
+one this VM already holds — restating a value the node would have supplied
+anyway grants nothing new. The same applies to `bridge_name` and
+`allowed_bridges`. The macvtap forwarding mode always comes
 from `[cvm.networking].macvtap_mode` and cannot be selected through deployment
 RPCs. These allowlists authorize attachment targets; an nwfilter is not a
 substitute for that authorization.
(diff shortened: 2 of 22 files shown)


Leechael:
Reviewed against `next` and read the netd/VMM paths end to end. Design and the on-host verification look right; the points below are the ones that would bite us on rollout.

**1. `max_net_queues` does not bound the default.**
`Networking::default_queue_pairs()` (`config.rs:1015`) is `vcpu.clamp(1, DEFAULT_QUEUE_SCALING_CAP)` and never reads `cvm.max_net_queues`; `validate_resolved_network` only checks the hard 64. With `max_net_queues = 4`, `--net-queues 5` is rejected while a 16-vCPU VM deployed without the flag gets 16 queue pairs. `vmm.toml:57-58` ("default to the VM's vCPU count, capped at this value") and the description say the opposite. Since `cvm.networking.queues` is rejected at config load, an operator has no node-level way to bound the default short of `vhost = false`. Suggest `vcpu.clamp(1, DEFAULT_QUEUE_SCALING_CAP.min(cfg.max_net_queues))`, which keeps the "raising the ceiling does not raise the default" intent.

**2. Queue-echo mismatch leaks the prepared interface; upgrade order is undocumented.**
In `prepare_filtered_networks` the rollback loop only runs on the `netd::request` error arm. The two post-prepare exits — missing macvtap `device` (`app.rs:638`) and the `queues` echo check (`app.rs:644`) — bail after netd has already created the interface, without `Remove` and before `set_runtime_networks`, so stop/remove never see it. Each rejected start leaves one TAP per NIC behind (plus the earlier NICs already in `prepared`).
The trigger is the skew you describe in the follow-ups: `dstack-netd.service` is a separate long-running process, so upgrading and restarting `dstack-vmm` alone leaves the old netd answering. On such a node every bridge (libvirt mode) or macvtap VM with more than one vCPU is refused at start until netd is restarted. Fail-closed is the right call; please route these two exits through the same rollback, and add an upgrade note (restart netd before or together with the VMM) to `docs/network-data-plane.md` — none of the docs mention it today.

**3. Existing nodes change their QEMU command line on upgrade with no config change.**
`vhost` is `Option<bool>` with `unwrap_or(true)`, so a vmm.toml that predates this PR turns vhost on. On a stock bridge node (`network_filter.mode = none`, no netd socket, helper at `/usr/lib/qemu/qemu-bridge-helper`) the netdev goes from `bridge,id=…,br=X` to `tap,id=…,br=X,helper=…,vhost=on`; QEMU then exits if the QEMU user cannot open `/dev/vhost-net`, and the VMM only checks that the node exists. Nodes running netd additionally move every multi-vCPU libvirt-bridge/macvtap VM to multiqueue on restart. I understand this is intentional; the description should state it as an upgrade step (verify `/dev/vhost-net` access as the QEMU user, or set `vhost = false` first), and I'd argue the safer default for a release is `vhost = false` with the on-by-default flip as a separate change.

**4. `update_networking` replaces `networks` wholesale.**
A request that changes `mode` without `queues`/`vhost` clears the previous per-VM tuning; a request with only tuning stores the node's current backend in the manifest. The UI round-trips the stored values so it is unaffected, but an RPC caller has to read-modify-write. Not a defect, but worth one sentence in the proto comments ("unset means cleared, not kept").

**Context that may be useful in the docs:**
- Runtime `vhost_net_start` failure does not exit: QEMU logs `falling back on userspace virtio` and continues on the userspace backend, and `vhostforce` does not change that. Worth listing next to the vhost thread-count check as something to alert on.
- The guest kernel has no `swiotlb=` and `CONFIG_SWIOTLB_DYNAMIC` off, so the bounce pool is a fixed `clamp(6% RAM, 64 MB, 1 GB)`. The default queue count follows vCPUs while the pool follows RAM; a 16 vCPU / 4 GB shape gets 246 MB. Your 8 vCPU / 8 GiB lab shape does not exercise it.

Separately, `port_map` on a non-user NIC is still accepted and silently dropped (only `hostfwd_index` consumes it). Not this PR's scope; I'll follow up.

Leechael:
Reference notes on how clouds pick a queue count — background for later tuning. dstack operators set `--net-queues` on the node or the VM; Phala Cloud will fill the same field from an instance type.

## 16 is a tuning convention, not a spec limit

| Layer | Cap | Source |
|---|---|---|
| virtio / Linux uapi | `VIRTIO_NET_CTRL_MQ_VQ_PAIRS_MAX = 0x8000` | [virtio_net.h](https://github.com/torvalds/linux/blob/master/include/uapi/linux/virtio_net.h) |
| Linux TAP (4.x+) | `MAX_TAP_QUEUES = 256` | [if_tap.h](https://github.com/torvalds/linux/blob/master/include/linux/if_tap.h) |
| TAP on 3.x | 8 | [Nova #1847367](https://bugs.launchpad.net/bugs/1847367) |
| RHEL 8/9 advice | one pair per vCPU, **up to 16** | [RHEL 9](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/monitoring_and_managing_system_status_and_performance/optimizing-virtual-machine-performance-in-rhel_monitoring-and-managing-system-status-and-performance), [RHEL 8](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/configuring_and_managing_virtualization/optimizing-virtual-machine-performance-in-rhel_configuring-and-managing-virtualization) |
| RHEL 7 | 1–256, no 16 cap | [RHEL 7](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/virtualization_tuning_and_optimization_guide/sect-virtualization_tuning_optimization_guide-networking-techniques) |
| QEMU cmdline | `queues=N`, `vectors=2N+2` | [linux-kvm Multiqueue](https://www.linux-kvm.org/page/Multiqueue) |

RHEL's "up to 16" is host-thread budget (one `vhost-net` worker per pair) plus diminishing returns. Same reason a default cap of 16 can sit below `max_net_queues`.

## How clouds encode it on the instance type

They publish **default per NIC**, **max per NIC**, and often an **instance-wide quota**. The SKU fills the hypervisor knob.

**AWS ENA** — per SKU three columns. General-purpose default plateaus at 8; network-optimized default goes to 16/32. Formula: `min(MAX_NUM_QUEUES_PER_ENI, vCPUs)`, where `MAX_NUM_QUEUES_PER_ENI` is 8 for most types and 32 for network-accelerated. Single ENI cannot exceed the instance vCPU count. 2025: queues are a pool allocatable across ENIs.

- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ena-queues.html
- https://github.com/amzn/amzn-drivers/blob/master/kernel/linux/ena/ENA_Linux_Best_Practices.rst
- https://aws.amazon.com/about-aws/whats-new/2025/05/aws-ec2-instances-ena-queue-allocation-network-interfaces/

**GCP** — formula from machine type + NIC driver, override at create.

- VirtIO: `min(vCPU / vNIC, 32)`
- gVNIC + Titanium: `min(vCPU / vNIC, 16)` (C4N 64; A2/G2 / some Tier_1 32)
- older gVNIC: `min(vCPU / vNIC / 2, 16)`
- total usually ≤ vCPU; Tier_1 can oversubscribe

- https://docs.cloud.google.com/compute/docs/network-bandwidth
- https://cloud.google.com/blog/products/compute/boost-your-vm-with-virtual-network-interfaces

**Aliyun ECS** — closest to a SKU column. Bind-to-type applies the default; API exposes `PrimaryEniQueueNumber`, `MaximumQueueNumberPerEni`, `TotalEniQueueQuantity`. U-instance is roughly `min(vCPU, 16)`. Stopped ENI can change within those two quotas; guest `ethtool -L` can lower further (lost on reboot).

- https://help.aliyun.com/zh/ecs/user-guide/nic-multi-queue
- https://help.aliyun.com/zh/ecs/user-guide/general-work-force

**OpenStack Nova** — `min(vcpus, max_queues)`. `max_queues` is operator config (legacy 1/8/256 by kernel major); flavor extra spec only enables/disables MQ.

- https://docs.openstack.org/nova/latest/admin/networking.html
- https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/libvirt-virtiomq.html

**Azure** — SKU capability is Accelerated Networking (SR-IOV), not virtio queue count. Different dataplane.

- https://learn.microsoft.com/en-us/azure/virtual-network/accelerated-networking-overview

## Practices that show up across those sources

Lined up with the numbers already in this PR (short connections 22k→6k conn/s from 1q to 8q; 64B UDP ~600k→3.0 Mpps). TDX makes the left-hand side worse than plain KVM: a cross-vCPU wakeup is an IPI plus a VM exit.

1. **One pair per vCPU is the useful maximum, not the useful default.** Extra pairs above vCPU are inert (guest `virtio_net` uses `min(vcpu, advertised)`). Too few pairs on a PPS-bound VM leave a single guest RX queue as the ceiling.
2. **Default and max are different columns.** AWS general-purpose defaults plateau at 8 on a 32-vCPU SKU; the max is 32. Aliyun's U-instance uses `min(vCPU, 16)` for both. Raising the request ceiling is not the same as retuning every VM's default.
3. **Workload picks the number, not vCPU count.**
   - many short TCP / RPC / conntrack churn → keep 1 (or a small N)
   - many concurrent flows / high PPS / large-UDP → raise toward `min(vcpu, 8)` and measure; 16 only if the host can afford the vhost threads
   - bandwidth with few large TSO flows saturates with a handful of pairs; more queues do not buy Gbps
4. **Compute-shaped SKUs stay conservative; network-shaped SKUs raise the default.** AWS states this directly (ENA best practices: more queues for network-intensive apps, fewer for CPU-intensive).
5. **Instance-wide quota once there is more than one NIC.** AWS / GCP / Aliyun cap Σ queues, not only per NIC.
6. **virtio + vhost-net costs more on the host than ENA / gVNIC / Azure VF.** Each pair is a `vhost-net` worker. A request ceiling of 32 can exist without the SKU default being 32.
7. **Who sets the knob.**
   - dstack operator: `--net-queues` / `max_net_queues` after measuring the VM
   - Phala Cloud: `default_queues` (and later `max_queues`) on the instance type; CreateVm sends that value. A plausible fill is `min(vcpu, family_default_cap)` — general/compute ≈ 8, a later network family ≈ 16

kvinwang:
Thanks for the review — every point landed. Status, in your order:

**1. `max_net_queues` not bounding the default** — fixed as specified: `default_queue_pairs` is now `vcpu.clamp(1, DEFAULT_QUEUE_SCALING_CAP.min(max_net_queues).max(1))`, with a test pinning that lowering the cap lowers the default while raising it only widens what a caller may request.

**2. Prepared-interface leak on the queue-echo mismatch** — fixed; both post-prepare exits now route through the same rollback as a failed prepare.

**3. Upgrade path** — resolved more conservatively than an upgrade note: the default is now **`vhost = false`**. An upgraded node builds byte-for-byte the same QEMU device it always did (userspace virtio, one queue pair) until the operator opts in; the vCPU-scaled queue default only applies once vhost is on. `docs/network-data-plane.md` gained an "Enabling vhost on a node" checklist (verify `/dev/vhost-net` access for the account QEMU runs under → restart netd with/before the VMM → rollback path), and the VMM now warns at startup when its own open of `/dev/vhost-net` is denied, not just when the node is missing.

**4. `update_networking` proto comment** — states it now: "Empty + update_networking=true resets to node default."

**Runtime fallback footnote** — verified on hardware (QEMU 8.2.2, the version dstack resolves on the test node, plus 10.2): an unopenable `/dev/vhost-net` kills QEMU at netdev init on every version — 8.2 actually SIGABRTs via [qemu#1486](https://gitlab.com/qemu-project/qemu/-/issues/1486)'s `assert(nc)` rather than exiting cleanly — with no userspace fallback. The `falling back on userspace virtio` path you pointed at is real but gated on `get_vhost_net()`, i.e. reachable only after the vhost fd opened successfully at launch, so an access problem never lands there. Both failure points are documented separately now, including that the runtime one is the single case where `GetInfo` can overstate the data plane.

**swiotlb footnote** — tested adversarially (dstack-0.6.0 guest, 16 vCPU × 16 queue pairs, offloads off, ~177 Gbit/s plus concurrent direct block I/O): not a risk at any realistic shape. The guest kernel clamps the pool to a 64 MB floor regardless of RAM — the small-RAM tail where the concern would live doesn't exist — and demand is ring-bounded at ~2 MB per queue pair; a deliberately undersized 32 MB pool produced zero `swiotlb buffer is full` events under full load. The only failure we could produce at all was the guest page allocator at 1 GB RAM × 16 vCPU — a RAM:vCPU ratio no deployment uses, and 2 GB at the same shape ran clean — recorded in the docs as a searchable symptom, not a knob. No `swiotlb=` parameter is warranted; it would only shrink the allocator that actually fails.

On your provider survey: with vhost now opt-in, no VM acquires the `min(vcpu, 16)` default silently on upgrade. Whether 16 should come down toward the 8-at-32-vCPU shape your data shows is still open — happy to lower `DEFAULT_QUEUE_SCALING_CAP` if you think the AWS column is the right anchor.

PR #1155, opened 2026-08-28
feat(netd): carry a VM's host port mappings to whoever configures the host

kvinwang:
Stacked on #1154, which is itself stacked on #1145. Review bottom-up.

## The bug

`port_map` is implemented as QEMU `hostfwd=` entries on a user-mode netdev:

```rust
let hostfwd_index = self.prepared.networks.iter()
    .position(|n| n.mode == NetworkingMode::User);
```

A bridge NIC has no user-mode netdev, so `hostfwd_index` is `None` and **not one `hostfwd=` is emitted**. There is no warning, no validation, and `GetInfo` keeps reporting the ports as though they worked. A VM moved from user mode to a bridge silently loses every published port.

I hit this while planning a production gateway's move off slirp: five port mappings, including the WireGuard endpoint 100+ CVMs depend on, would have evaporated on the first boot with nothing in the log to say so.

## Why the VMM cannot just fix it

It runs without `CAP_NET_ADMIN` by design — that is the entire reason `netd` exists. And the userspace forwarder that used to cover this was removed in e2e607fd4, correctly: proxying on the host hands back the per-packet cost that leaving user mode was meant to escape.

`netd` is privileged, and it is the only component that sees every VMM instance on a host — so it is also the only one that can arbitrate a host port between them. This host runs three dstack deployments against one netd.

## What this does

Carries the requirement instead of implementing it. `ingress` on `prepare_bridge` states protocol, host address, host port, guest port; the response reports what was established plus the `guest_ip` netd assigned.

**Every field is caller-named.** That is what `bridge`, `mac` and `queues` already get, and the opposite of `filtered`. The line is not "policy vs. mechanism" — it is whether netd can check what it is handed. An nwfilter name cannot be checked for whether it filters anything (`allow-arp` satisfies "some filter" and drops nothing), so naming one is excluded. A host port is a closed, enumerable space a policy can be stated over. Naming is not deciding: which ports may be handed out stays netd's configuration, exactly as `allowed_bridges` governs the bridge a caller names.

`host_address` is not decoration — an admin port on loopback and a published one differ only there.

**The netd in this repo does not forward.** It does not name `ingress` in `hello`, and it refuses a request carrying ports rather than building the interface without them — refused alongside the other checks about the request, before the host is inspected, since it is a fact about the request. Accepting ports and forwarding nothing is the exact failure being ended.

**The VMM asks `hello` first** and does not send ports to a netd that will not forward them; it warns instead. A VM deployed before any of this has been running with its ports dropped, and refusing to launch it now would turn a silent misconfiguration into an outage on upgrade. `macvtap` and `custom` get the same warning permanently — the first bypasses the host bridge, the second hands the netdev string to the operator.

**`GetInfo` grows `guest_ip` and `ingress` per interface.** What the node did, against the `ports` on the configuration, which is what was asked for. An empty `ingress` against a non-empty request is now a visible difference instead of silence.

Reporting `guest_ip` also closes what `ReportDhcpLease` was for, without reinstating a DHCP hook: a netd that forwards is necessarily also the authority on the address, because a DNAT rule needs one at install time and a lease does not exist until the guest has booted.

## Compatibility

Additive. `ingress` defaults empty, so an older caller asks for nothing and an existing bridge VM behaves exactly as before apart from a new warning. Proto fields are new tags; `ports` on the configuration is untouched.

## Testing

`cargo test -p dstack-vmm` — 190 pass, 3 new. `cargo clippy -- -D warnings` clean.

New cases cover the request shape and the empty default, the refusal firing before host inspection (and *not* firing when nothing is asked for), and `hello` not advertising a feature this netd lacks.

One of these caught a bad assertion of mine: I had asserted on a bridge named `br0` not existing, and the dev host has a `br0`. The test now uses a name nothing will have.

Not covered: an actual forwarding netd, which by construction is not in this repository.

### DIFF
--- a/docs/bridge-networking.md
+++ b/docs/bridge-networking.md
@@ -199,6 +199,41 @@ The remaining bytes are derived from the VM ID hash. The prefix applies to all n
 - Docker's nftables chains (`DOCKER-FORWARD`) run before libvirt's but do not block virbr0 traffic
 - Use `setup-bridge.sh check --bridge <name>` to diagnose missing rules
 
+### Host port mappings
+
+A VM's `port_map` has always been implemented as QEMU `hostfwd=` entries on a
+user-mode netdev. A bridge NIC has no such netdev, so **the VMM does not forward
+host ports for bridge VMs** — it is deliberately run without `CAP_NET_ADMIN`,
+and a userspace proxy on the host would give back the per-packet cost that
+moving off user mode was meant to escape.
+
+What the VMM does instead is carry the requirement to `netd`, which is
+privileged and is the only component that sees every VMM instance on the host —
+and therefore the only one that can arbitrate a host port between them. Whether
+a given `netd` implements forwarding is its own business; the one in this
+repository does not, and says so through `hello` rather than accepting ports it
+will not forward.
+
+So on a node whose `netd` does not forward:
+
+- the VM starts normally and its bridge networking works
+- its port mappings do not apply, and the VMM says so in its log at every launch
+- `GetInfo` reports an empty `ingress` on the interface, against the non-empty
+  `ports` on the configuration — that difference is the signal
+
+Expose such a VM by reaching its address on the bridge directly, by putting an
+L7 proxy in front of it, or by running a `netd` that forwards. `GetInfo` reports
+the interface's `guest_ip` when `netd` is the authority for addresses on that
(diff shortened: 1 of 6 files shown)


kvinwang:
Folding into #1154 — the two together grew well past what was asked for (host port info plus workdir on the netd API). #1154 is being rewritten to just that.

PR #1162, opened 2026-09-02
fix: fix install script informational echo's

WimYedema:
The resolve_source function echo's a return value but also informational messages. I created an `info` function for the informational messages such that the return value is not muddied. For completeness-sake I also added an `error` function.

### DIFF
--- a/dstack/scripts/install.sh
+++ b/dstack/scripts/install.sh
@@ -238,7 +238,7 @@ fi
 need_cmd cargo
 need_cmd install
 
-checkout=$(resolve_source)
+checkout=$(resolve_source | tail -1)
 core_checkout=$(core_dir "$checkout")
 bin_dir="$prefix/bin"
 
--- a/dstack/scripts/install.sh
+++ b/dstack/scripts/install.sh
@@ -58,11 +58,11 @@ cleanup() {
 trap cleanup EXIT INT TERM
 
 info() {
-    echo "$*" >&2
+    printf '%s\n' "$*" >&2
 }
 
 error() {
-    echo "error: $*" >&2
+    printf 'error: %s\n' "$*" >&2
 }
 
 while [ "$#" -gt 0 ]; do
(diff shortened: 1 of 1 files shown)


kvinwang:
Pushed a follow-up commit to this branch (`9e1b99d`) — hope you don't mind, `maintainerCanModify` was on.

### The stdout leak isn't only from the script's own `echo`s

`resolve_source()` returns the checkout path on stdout, so *anything* written there lands in `$checkout`. git writes to stdout too:

- `git checkout <branch>` prints `Switched to branch`/`Already on` to **stderr**, but the branch-tracking line `Your branch is up to date with 'origin/next'.` to **stdout**.
- `git pull --ff-only` prints its `Updating ..`/`Fast-forward` summary to **stdout**.

So the branch as submitted still broke, including on the default `curl | sh` path (`--ref` defaults to the `next` branch; a tag/SHA happens to be fine because detached-HEAD messages all go to stderr):

```
$ sh install.sh --src /tmp/e2esrc --repo ... --ref next --prefix ... --no-sudo
cloning dstack source into /tmp/e2esrc
Cloning into '/tmp/e2esrc'...
Already on 'next'
building dstackup from Your branch is up to date with 'origin/next'.
/tmp/e2esrc
install.sh: 268: cd: can't cd to Your branch is up to date with 'origin/next'.
/tmp/e2esrc
exit=2
```

The fix redirects the git subshells and the `git clone`s as a whole (`) >&2`) rather than adding `-q` per command, so a git call added later can't reintroduce the same bug.

### Copilot's `printf` note

Applied — it's a real difference, not just style. Under `dash`, `info "-n"` with `echo` swallows the line entirely:

```
$ dash -c 'info() { echo "$*" >&2; }; info "-n"; info "-e whatever"'
-e whatever          # "-n" vanished, "-e" consumed as a flag
$ dash -c 'info() { printf "%s\n" "$*" >&2; }; info "-n"; info "-e whatever"'
-n
-e whatever
```

Also routed `building dstackup from ...` through `info` for consistency; the final install summary stays on stdout.

### Verification

Harness that truncates the script after `resolve_source()` and checks that stdout is exactly the path, across all six resolution paths:

| Test | base (pre-PR) | this PR before | after |
|---|---|---|---|
| temp clone (`curl \| sh`) | FAIL | FAIL | PASS |
| `--src` fresh clone | FAIL | FAIL | PASS |
| `--src` update (`git pull`) | FAIL | FAIL | PASS |
| `--src --ref <tag>` | FAIL | PASS | PASS |
| cwd is a checkout | PASS | PASS | PASS |
| `script_checkout` via `$0` | PASS | PASS | PASS |

Plus a full end-to-end run with a stubbed `cargo` (clone → build → `install_bin` → summary): exits 0, binary installed, stdout carries only the final summary, all progress on stderr. `prek run --from-ref <base> --to-ref HEAD` passes, shellcheck included.

PR #1178, opened 2026-09-05
feat(guest-agent): expose GPU telemetry via GpuInfo RPC

Leechael:
## Why

Operators and the cloud control plane need to know whether a GPU attached to a CVM is actually being used: SM utilization, framebuffer usage, temperature, power, and whether the confidential-computing handshake has completed. None of that is visible from the host. With VFIO passthrough the host driver never binds the card, so host-side NVML and DCGM cannot read it, and in CC mode the device is attested to the guest only. The only vantage point is inside the guest, which is where `dstack-guest-agent` already runs. This PR gives the agent a GPU telemetry surface.

## Design

The first draft put a `gpus` list on `SystemInfo` and called `Nvml::init()` on every request. Review found three problems, and a survey of how DCGM exporter, go-nvml, nvml-wrapper, the Kubernetes device plugin, and the GCP/AWS in-VM agents handle this confirmed the direction:

- **One NVML handle per process.** go-nvml and nvml-wrapper both document init-once, shutdown-once. nvml-wrapper's `Nvml` shuts the library down on `Drop`, so a request-scoped handle meant init+shutdown on every `/metrics` scrape.
- **GPU collection must not share fate with SysInfo.** DCGM exporter's known failure mode is one GPU in `ERR!` state taking the whole exporter down. Here GPU collection had SysInfo's 20 s timeout and `spawn_blocking`; a stuck NVML call (GPU reset, driver assert, CC handshake) would stall host and guest health reporting.
- **Missing must be distinguishable from zero.** The draft swallowed init and per-field errors, so "no GPU", "NVML down", "query failed", and a genuine 0 all looked the same. Two fields (`memory_reserved_bytes`, `memory_usable_bytes`) were placeholders with fabricated values.
- **Proto changes cross the VMM.** `ProxiedGuestApi` decodes and re-encodes, so a `SystemInfo` field the VMM does not know is dropped silently. The VMM must ship the same proto.

What this PR does instead:

- Independent `GpuInfo` RPC on `GuestApi` and `ProxiedGuestApi` (named like `Info` / `SysInfo` / `NetworkInfo`). `SystemInfo` is unchanged.
- Process-wide NVML handle in `gpu_info.rs`. A failed init is cached for 60 s so driver-less images do not `dlopen` on every scrape.
- Sampling is serialized with `try_lock`. The 5 s RPC timeout does not cancel a stuck NVML call; later callers receive the last snapshot, or `error: "sampling in progress"` if there is none, instead of queueing behind it. Successful snapshots are cached for 5 s and shared by the RPC, `/metrics`, and the dashboard.
- Every numeric field is `optional`. Per-device `error` lists the queries that failed. `NotSupported` logs at debug, because CC mode disables some counters by design; other per-field errors warn once per (device, field).
- Identity is `uuid` (stable), with `index` and `pci_bus_id` (`nvmlPciInfo.busId`) alongside, matching the DCGM exporter label set so host-side tooling that speaks BDF can correlate.
- `cc_ready` (system-wide `nvmlDeviceGetConfComputeGpusReadyState`) and per-device `cc_enabled`, so "driver present but still attesting" is distinguishable from "NVML unavailable".
- The two placeholder memory fields are removed; nvml-wrapper 0.12.1 has no `memory_info_v2`.

## Where the data is exposed, and who can see it

| Surface | Path | Audience | Gate |
|---|---|---|---|
| vsock `GuestApi.GpuInfo` | guest agent, vsock port 8000 | the VMM only | none; host operator channel |
| VMM `ProxiedGuestApi.GpuInfo` | `POST /guest/GpuInfo?json` on the VMM socket | host operators, control plane | VMM access control |
| `/metrics` on the guest's external port | Prometheus text, `dstack_gpu_*` series | anyone who can reach the port | `public_sysinfo` in app-compose |
| Dashboard | GPU table on the guest dashboard | anyone who can reach the port | `public_sysinfo` in app-compose |

The public surfaces reuse the existing `public_sysinfo` switch. `/metrics` already refused requests when it is off; the dashboard GPU section is wrapped in the same check. An app owner who keeps `public_sysinfo: false` exposes nothing new, and the host still gets telemetry through the VMM.

Metrics series: `dstack_gpu_nvml_up`, `dstack_gpu_cc_ready`, and per device `dstack_gpu_utilization_percent`, `dstack_gpu_memory_utilization_percent`, `dstack_gpu_memory_{total,used,free}_bytes`, `dstack_gpu_temperature_celsius`, `dstack_gpu_power_usage_milliwatts`, `dstack_gpu_query_errors`, all labeled `index`, `uuid`, `pci_bus_id`. A field that failed to sample emits no series rather than a 0.

## Tests

- Proto round trip preserves `None` vs `Some(0)` and an unset `cc_ready`.
- Collector does not panic without NVML; unavailability is reported as `error` with an empty device list, not as a silent empty success. The test passes with or without a driver present.

## Validation

- `cargo build -p dstack-guest-agent -p dstack-vmm`, `cargo test -p dstack-guest-agent -p guest-api`, `cargo fmt --check`, `cargo clippy -p dstack-guest-agent -p dstack-vmm -- -D warnings` on a TDX lab host without GPUs.
- A B200 test image was built from `next` + #1181 + #1177 + this branch (os_image_hash `a7a7646b0111236340b576645a4c8ff29c89facf8510c1019e52383f675326a7`, `git_revision` clean) and the rootfs was checked offline for the NVML symbols and the 595.91.07 driver. Live single-GPU validation on that image, with a VMM built from the same tree, is the remaining step.

## Known limits

- XID event based health tracking is out of scope.
- The exact set of NVML fields that return `NotSupported` under CC mode on B200 has not been enumerated yet; the design tolerates it but the list should be recorded once the live run happens.
- The VMM must be released with the same proto for `/guest/GpuInfo` to exist.

## References

Prior art surveyed for the design decisions above:

- NVIDIA DCGM exporter: https://github.com/NVIDIA/dcgm-exporter , metric and label reference https://docs.nvidia.com/datacenter/dcgm/latest/reference/dcgm-exporter-metrics.html , docs https://docs.nvidia.com/datacenter/cloud-native/gpu-telemetry/latest/dcgm-exporter.html , one bad GPU aborting the exporter https://github.com/NVIDIA/dcgm-exporter/issues/46
- go-nvml, init-once lifecycle: https://github.com/NVIDIA/go-nvml
- nvml-wrapper (Rust), process-wide handle guidance: https://github.com/rust-nvml/nvml-wrapper , https://docs.rs/nvml-wrapper
- nvidia_gpu_exporter, per-scrape vs background collection and graceful degradation: https://github.com/utkuozdemir/nvidia_gpu_exporter , https://github.com/utkuozdemir/nvidia_gpu_exporter/blob/master/METRICS.md
- Kubernetes device plugin, UUID/BDF identity and per-device XID health: https://github.com/NVIDIA/k8s-device-plugin , https://github.com/NVIDIA/k8s-device-plugin/blob/main/docs/gpu-feature-discovery/README.md , https://github.com/NVIDIA/k8s-device-plugin/issues/1014
- In-VM cloud agents: Google Ops Agent GPU metrics https://docs.cloud.google.com/compute/docs/gpus/monitor-gpus , https://docs.cloud.google.com/monitoring/api/metrics_opsagent ; AWS CloudWatch agent NVIDIA metrics https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Solution-NVIDIA-GPU-On-EC2.html ; Azure guidance (Telegraf + nvidia-smi / DCGM) https://techcommunity.microsoft.com/blog/azurehighperformancecomputingblog/comprehensive-nvidia-gpu-monitoring-for-azure-n-series-vms-using-telegraf-with-a/4257402
- NVIDIA confidential computing, counters disabled and CC ready state: https://developer.nvidia.com/blog/confidential-computing-on-h100-gpus-for-secure-and-trustworthy-ai/ , https://cloud.google.com/confidential-computing/confidential-vm/docs/verify-confidential-computing-mode-on-gpu , https://cacm.acm.org/practice/creating-the-first-confidential-gpus/

### DIFF
--- a/docs/security/cvm-boundaries.md
+++ b/docs/security/cvm-boundaries.md
@@ -35,7 +35,7 @@ This is the main configuration file for the application in JSON format:
 | local_key_provider_enabled | 0.3.1 | boolean | Use a local key provider |
 | key_provider_id | 0.5.1 | string | Optional pin for the key provider identity (hex-encoded bytes). For `kms` this is the KMS CA public key; for `local` the sealing-provider MR. For `tpm` and `none` it must be an empty string — the TPM app-root public key is instance-specific and is not used as a provider id or measured as one. |
 | public_logs | 0.3.3 | boolean | Whether logs are publicly visible |
-| public_sysinfo | 0.3.3 | boolean | Whether system info is public |
+| public_sysinfo | 0.3.3 | boolean | Whether system info is public. Covers the guest dashboard and `/metrics`, including the `dstack_gpu_*` series (since 0.6.0), which expose each GPU's UUID and PCI bus address alongside utilization, memory, temperature and power. |
 | public_tcbinfo | 0.5.1 | boolean | Whether TCB info is public |
 | allowed_envs | 0.4.2 | array of string | List of allowed environment variable names |
 | no_instance_id | 0.4.2 | boolean | Disable instance ID generation |
--- a/dstack/Cargo.lock
+++ b/dstack/Cargo.lock
@@ -2024,6 +2024,7 @@ dependencies = [
  "listenfd",
  "load_config",
  "nvattest",
+ "nvml-wrapper",
  "or-panic",
  "ra-rpc",
  "ra-tls",
--- a/dstack/Cargo.lock
+++ b/dstack/Cargo.lock
@@ -2002,6 +2002,7 @@ dependencies = [
  "anyhow",
  "base64 0.22.1",
  "bollard",
+ "cached-cell",
  "cc-eventlog",
  "cert-client",
  "chrono",
@@ -2023,8 +2024,8 @@ dependencies = [
  "libc",
  "listenfd",
  "load_config",
+ "lspci",
  "nvattest",
- "nvml-wrapper",
  "or-panic",
  "ra-rpc",
  "ra-tls",
@@ -2292,11 +2293,13 @@ dependencies = [
  "ez-hash",
  "fs-err",
  "getrandom 0.3.4",
+ "guest-api",
  "hex",
  "hex_fmt",
  "host-api",
  "k256",
  "libc",
+ "lspci",
  "luks2",
  "nvattest",
  "nvml-wrapper",
@@ -4292,6 +4295,7 @@ version = "0.6.0"
 dependencies = [
  "anyhow",
  "insta",
+ "tempfile",
 ]
 
 [[package]]
--- a/dstack/dstack-util/Cargo.toml
+++ b/dstack/dstack-util/Cargo.toml
@@ -42,6 +42,7 @@ tpm-attest.workspace = true
 tpm2.workspace = true
 tpm-qvl = { workspace = true, features = ["crl-download"] }
 host-api = { workspace = true, features = ["client"] }
+guest-api.workspace = true
 cmd_lib.workspace = true
 toml.workspace = true
 dcap-qvl.workspace = true
@@ -60,6 +61,7 @@ sodiumbox.workspace = true
 libc.workspace = true
 luks2.workspace = true
 nvml-wrapper.workspace = true
+lspci.workspace = true
 scopeguard.workspace = true
 tempfile.workspace = true
 ez-hash.workspace = true
--- a/dstack/dstack-util/src/gpu_info.rs
+++ b/dstack/dstack-util/src/gpu_info.rs
@@ -1,0 +1,282 @@
+// SPDX-FileCopyrightText: © 2025 Phala Network <dstack@phala.network>
+//
+// SPDX-License-Identifier: Apache-2.0
+
+//! One-shot NVML sampler behind `dstack-util gpu-info`.
+//!
+//! NVML calls cannot be cancelled: a driver that wedges during a GPU reset or a
+//! CC handshake blocks the calling thread until it decides to return. Running
+//! the sample in a short-lived process lets `dstack-guest-agent` kill it and
+//! move on, and keeps `libnvidia-ml` out of the long-lived agent entirely.
+//!
+//! One process per sample also means `Nvml::init()` runs fresh every time, so a
+//! driver that loads after the agent started is picked up on the next sample
+//! instead of being cached as "no GPU" for the agent's lifetime.
+//!
+//! This command owns collection only. Caching, presence gating and timeouts
+//! belong to the caller.
+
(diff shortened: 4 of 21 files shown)


kvinwang:
Ran this branch on real hardware — an H200 SXM 141 GB (`19:00.0`) in CC mode, inside a TDX CVM, with a VMM built from the same tree merged with #1065 for GPU assignment. Posting the results here rather than editing the description, since the three items under **Known limits** now have answers and the **Validation** section's remaining step is done.

## The three unknowns

**The set of NVML fields that return `NotSupported` under CC mode is empty.** At least on a single H200: `query_errors = 0`, stderr silent, SM utilization, all three memory fields, temperature and power all available, with `cc_enabled = 1` and `cc_ready = 1`. The design tolerating `NotSupported` is still right — it just never fires here. B200 remains unmeasured.

**One sample takes 94-99 ms**, fork to exit, including `dlopen`, `nvmlInit_v2`, enumeration, every field query and JSON serialization. Corroborated independently: two scrapes 20 s apart reported ages of 14.41 s and 19.96 s, putting the refresh the first one triggered at ~40 ms. So a cold `nvmlInit_v2` under CC is not the expensive thing it was reasonable to assume.

**The zero-cost claim for GPU-less guests holds on real hardware** once the PCI gate is in: 0 collector processes, 0 NVIDIA devices on PCI, across repeated scrapes.

Incidentals: `memory_total_bytes = 150754820096` (140.40 GB) matches an H200 SXM 141 GB, and the UUID and PCI BDF formats line up with the DCGM exporter label set, so host-side correlation works as intended.

## One thing the live run found

`/metrics` returns no GPU data at all on a healthy card. The non-blocking path serves a placeholder whenever the snapshot is older than the 5 s TTL, and at any realistic Prometheus interval every scrape is older than that — so the endpoint reports `dstack_gpu_nvml_up 0` and zero device series on a working H200. The two ages above, 14.41 s and 19.96 s, are exactly the case: both were placeholders before the fix, both return data after.

Full nine-check matrix, plus fixes for that and a few smaller things (a GPU-less guest forking a resident helper on every scrape, `cc_enabled` being system-wide rather than per-device, `error` not being safely countable, and a wedged driver being respawned on the success cadence) are in #1184, stacked on this branch so it shows only the added commits. Happy to fold any of it into this PR instead if you would rather land one change.

PR #1193, opened 2026-09-09
feat(vmm): add a status filter to the console instance list

Leechael:
## Problem

The VMM console can search instances by keyword, but not by lifecycle status. The list is paginated server-side, so a client-only filter would only cover the current page and would report the wrong total.

## Fix

- Add `StatusRequest.status` (`running`, `stopped`, `exited`, `stopping`, `removing`). Empty keeps the previous all-status behavior.
- Apply the filter in `list_vms` before pagination, using the same status values already shown in the Status column.
- Share status computation through `vm_runtime_status` so the filter and the list agree.
- Add a toolbar select next to search. Changing it resets to page 1, matching page-size changes.

Old clients omit the field, so existing `Status` callers are unchanged.

## Verification

- `npm run build` in `dstack/vmm/ui`
- `cargo check -p dstack-vmm-rpc`
- `prek` on the changed files


kvinwang (approved):
LGTM

PR #1204, opened 2026-09-12
fix(vmm): detect qemu version before each vm start

Leechael:
## Problem

The VMM detects the QEMU version once at config load and caches it in `Config.cvm.qemu_version`. Every `vm_config` generated afterwards copies the cached value, while the QEMU actually launched is whatever binary currently sits at `qemu_path`. If the binary is replaced while the VMM keeps running (e.g. a package upgrade or a manual swap), the declared version goes stale and no longer matches the binary being executed.

The verifier picks an ACPI table model from the declared `qemu_version`, so a stale declaration makes RTMR0 replay fail and the CVM unattestable. Observed in production: running VMs on a host were executing `/usr/bin/qemu-system-x86_64` 9.2.1 while their host-shared `.sys-config.json` declared `qemu_version: 10.1.0`, cached from before the binary was replaced. Restarting the VMs refreshed the declaration to 9.2.1 and full quote/event-log verification passed again.

## Fix

Detect the version before every VM start instead of once at config load, and make the `vm_config` written for a start use that same detection result:

- `CvmConfig::resolve_qemu_version()` returns the explicit `qemu_version` from the config file when one is set (a serde-skipped flag records explicitness at config load, so the override contract is unchanged); otherwise it re-runs `qemu --version` against the binary currently at `qemu_path`, logging when the version changed since config load.
- `sync_dynamic_config()` — the single entry point every start funnels through (RPC `StartVm`, auto-start on VMM boot, and the auto-restart task) — resolves the version and threads it through `make_sys_config()` into the `vm_config` it writes for that start.
- If the per-start re-detection fails, it falls back to the load-time value with a warning, mirroring the previous "start without a fresh version" behavior rather than failing the start.
- One-shot mode keeps passing the load-time value: it loads its config immediately before the start, so the value is already fresh and its behavior is unchanged.

This does not bind the declared version to the executed binary atomically — the binary can still be replaced between detection and exec. It removes the staleness window of a long-running VMM, which is the mismatch observed in production.

## Verification

- New regression test `qemu_version_is_redetected_after_binary_replacement`: a fake `qemu-system-x86_64` reporting 10.1.0 is "detected at config load", then replaced with one reporting 9.2.1. Before the fix the next start still declares the stale 10.1.0 (the test fails with `left: Some("10.1.0"), right: Some("9.2.1")`); after the fix it declares 9.2.1.
- New tests `explicit_qemu_version_overrides_detection` (explicit config wins even when the binary reports a different version) and `qemu_version_detection_failure_falls_back_to_load_time_value` (detection failure falls back to the load-time value, then to `None`).
- New test `vm_config_declares_the_given_qemu_version` covering the `make_vm_config` threading: the passed version is what lands in `vm_config`.
- `cargo test -p dstack-vmm`: 132 passed, 0 failed.
- `cargo clippy -- -D warnings -D clippy::expect_used -D clippy::unwrap_used --allow unused_variables` (the CI invocation): clean.
- `cargo fmt --all -- --check` and the prek hooks: clean.

### DIFF
--- a/dstack/vmm/src/app.rs
+++ b/dstack/vmm/src/app.rs
@@ -1305,12 +1305,17 @@ impl App {
         let mr_config = work_dir
             .prepare_mr_config(&manifest, &cfg.cvm, &app_compose)
             .context("Failed to prepare mr_config")?;
+        // Detect the version of the QEMU binary this start will execute so
+        // the declared version cannot go stale if the binary is replaced
+        // while the VMM keeps running.
+        let qemu_version = cfg.cvm.resolve_qemu_version();
         let sys_config_str = make_sys_config(
             cfg,
             &manifest,
             &hex::encode(compose_hash),
             mr_config,
             app_compose.requirements.as_ref(),
+            qemu_version,
         )?;
         fs::write(shared_dir.join(SYS_CONFIG), &sys_config_str)
             .context("Failed to write vm config")?;
@@ -1603,6 +1608,7 @@ pub(crate) fn make_sys_config(
     compose_hash: &str,
     mr_config: Option<String>,
     requirements: Option<&dstack_types::Requirements>,
+    qemu_version: Option<String>,
 ) -> Result<String> {
     let image_path = cfg.image.path.join(&manifest.image);
     let image = Image::load(image_path).context("Failed to load image info")?;
@@ -1644,6 +1650,7 @@ pub(crate) fn make_sys_config(
         compose_hash,
         mr_config.clone(),
         requirements,
+        qemu_version,
     )?;
     let mut sys_config = json!({
         "kms_urls": kms_urls,
@@ -1723,6 +1730,7 @@ fn make_vm_config(
     _compose_hash: &str,
     mr_config: Option<String>,
     requirements: Option<&dstack_types::Requirements>,
+    qemu_version: Option<String>,
 ) -> Result<serde_json::Value> {
     let platform = cfg.cvm.resolved_platform();
     let is_amd_sev_snp = platform == crate::config::CvmPlatform::AmdSevSnp && !manifest.no_tee;
@@ -1804,7 +1812,7 @@ fn make_vm_config(
         memory_size: manifest.memory as u64 * 1024 * 1024,
         qemu_single_pass_add_pages: cfg.cvm.qemu_single_pass_add_pages,
         pic: cfg.cvm.qemu_pic,
-        qemu_version: cfg.cvm.qemu_version.clone(),
+        qemu_version,
         pci_hole64_size: cfg.cvm.qemu_pci_hole64_size,
         hugepages: manifest.hugepages,
         num_gpus: gpus.gpus.len() as u32,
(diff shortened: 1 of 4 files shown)


kvinwang:
I pushed two commits on top of yours (`d20d66f`, `e974c56`). The fix is the same: declare the version of the binary each start actually runs. The implementation is much smaller, and one behavior changes.

### Why it could be simpler

The explicit flag, the load-time fallback and the new `qemu_version` parameter all had one cause. Config load wrote the detected version back into `cvm.qemu_version`, so that field meant "configured *or* detected". Once that write-back is gone, all three can go too.

**`d20d66f` refactor(vmm): resolve the qemu version where vm_config is built**

- Removed the version detection from `Config::extract_or_default`. `qemu_version` is now only an explicit override, so `qemu_version_explicit` is gone.
- `make_vm_config` calls `cfg.cvm.resolve_qemu_version()` directly. It already takes `cfg`, so no signature changes. The `None` arguments added at 9 test call sites are gone, and `one_shot.rs` needs no special case.
- **Behavior change:** if the version can't be detected and isn't configured, the start now fails with `set `qemu_version` in vmm.toml to declare it explicitly`. Falling back to the load-time value would declare the stale version this PR is removing. Declaring nothing would make the verifier use a default ACPI model and boot a CVM that can't be attested. Either way the host is misconfigured, and it's easier to diagnose at start than after boot. `vm_config.qemu_version` is now always `Some`.
- The resolved version is logged at `info` on each start.

**`e974c56` fix(vmm): read the qemu version past a wrapper's banner**

Now that failed detection blocks the start, the parser needs to be reliable:

- It searches every line of stdout **and** stderr for `QEMU emulator version X`. Before, it read only the first line of stdout, so a wrapper script that printed a banner first failed detection.
- It no longer guesses a version-shaped word when that wording is missing. With every line searched, a guess could pick up a banner like `exec /usr/libexec/qemu-kvm-8.2.0` and return `8.2.0`. A wrong version is worse than an error. On failure, the error includes a short excerpt of the actual output.
- `test_parse_qemu_version_fallback` becomes `test_parse_qemu_version_without_qemu_wording` and now expects an error. The Debian, old-Debian, simple, rc and invalid parse tests pass unchanged. Mainline QEMU prints this string as a literal and distro packaging only changes the text in parentheses, so real QEMU binaries are not affected.

### Tests

Net result vs. `b20c00f`: +136/−62 across 3 files (was +193/−10 across 4).

| Test | Covers |
|---|---|
| `qemu_version_follows_the_binary_at_qemu_path` | the production regression (10.1.0 → replaced binary → 9.2.1) |
| `explicit_qemu_version_wins_over_the_binary` | explicit override |
| `a_wrapper_banner_does_not_hide_the_version` | banner on stderr, version on a later line |
| `an_undetectable_qemu_version_fails_the_start` | a version-shaped banner is not guessed, and the error points at `qemu_version` |

`vm_config_declares_the_given_qemu_version` is removed because the parameter it tested no longer exists. The two `app.rs` test config helpers now set `qemu_version`, since no QEMU binary is present under test.

`cargo test -p dstack-vmm --all-features`: 132 passed. The CI clippy command (`-D warnings -D clippy::expect_used -D clippy::unwrap_used`) and `cargo fmt --check` are clean. I checked each commit on its own.

### For the PR description

- **Breaking:** a host whose `qemu_path` doesn't print `QEMU emulator version` (a wrapper or custom build) now refuses to start VMs until `qemu_version` is set. Before, the VM booted but couldn't be attested.
- A misconfigured `qemu_path` now shows up on the first VM start, not as a warning when the VMM starts. I left out a startup pre-check on purpose: detection succeeding at startup doesn't mean it will succeed at VM start, which is the problem this PR fixes.

PR #1225, opened 2026-09-15
fix(yocto): avoid preserving source groups during unpack

WimYedema:
The dstack recipes copy source trees from outside BitBake's work directory. Archive-mode rsync attempts to preserve their group ownership, which can fail when the build user cannot assign the source group.

Disable group preservation while retaining the other archive-mode attributes.


kvinwang:
Right fix, right scope — these two recipes are the only in-tree places staging a tree from outside `WORKDIR`, and both `do_install`s use explicit `install -m`, so dropping ownership on the staging copy can't reach the packaged image.

One gap, so I pushed a follow-up (fdfdfbd) instead of a round-trip: `--no-group` only helps while rsync believes it is *not* super-user. Under rootless docker/podman or `sudo os/build.sh`, archive mode still applies `-o` and fails the same way — `chown ... Invalid argument (22)`, exit 23 — and the aborted transfer leaves the file at the temp-file mode (`0600`) rather than the source mode, so the staged tree goes silently wrong. Verified with rsync 3.2.7: only `--no-owner --no-group` survives all three privilege setups.

Ownership is meaningless for a `WORKDIR` staging copy, so the follow-up drops both and adds a short comment above each `do_unpack` recording why, otherwise this folds back into a bare `-a` at the next cleanup. `-p` is kept deliberately — executable bits still matter.

LGTM with that on top.

PR #1232, opened 2026-09-20
fix(guest-agent): reject a certificate validity bound it cannot represent

kvinwang:
## The bug

`GetTlsKey` and the v1 `IssueCert` take `not_before` and `not_after` from the request and hand them to certificate issuance unchecked. The conversion is

```rust
fn unix_time_to_system_time(secs: u64) -> SystemTime {
    UNIX_EPOCH + Duration::from_secs(secs)
}
```

which panics on overflow, and the workspace builds release binaries with `panic = "abort"`. So one request kills the agent:

```console
$ curl --unix-socket /var/run/dstack.sock -X POST -H 'Content-Type: application/json' \
    -d '{"subject":"localhost","not_after":18446744073709551615}' http://localhost/GetTlsKey
```
```
thread 'rocket-worker-thread' panicked at library/std/src/time.rs:614:31:
overflow when adding duration to instant
Aborted
```

The guest socket is bind-mounted into application containers by design, so any container in the CVM can take down key provisioning, quote generation and log aggregation for the whole guest, permanently, with one call. The same bound reaches the KMS through `SignCert` — `CaCert::sign_csr` converts `CertConfigV2.not_after` the same way — so an authorized CVM can abort the KMS every other tenant depends on.

The existing `validate_cert_validity` only compares the two bounds against each other, so it passes whenever one of them is absent or smaller than the other. `{"not_before": 0, "not_after": u64::MAX}` walks straight through it.

## The fix

Bound both fields against the last instant RFC 5280 can encode (9999-12-31T23:59:59Z) and report which field was out of range, so the caller gets an actionable error rather than a dead socket. Make the conversion in `ra-tls` fallible too, so the library cannot be made to panic through any other caller — `SignCert` reaches it without passing through the guest agent's validator.

## Verification

Against a simulator built from this branch:

| `not_after` | before | after |
|---|---|---|
| `18446744073709551615` | process `Aborted` | `400 not_after must not be later than 253402300799` |
| `9223372036854775807` | process `Aborted` | `400 not_after must not be later than 253402300799` |
| `253402300800` | process `Aborted` | `400 not_after must not be later than 253402300799` |
| `253402300799` | — | `200`, two-certificate chain |

`not_before` and the v1 `/v1/IssueCert` mount behave identically. The listener answers `Info` with `200` after every row.

The bug was found by an adversarial request-contract sweep over the guest agent's 29 indexed RPC methods: 1016 vectors, 4 of which aborted the process, all four this root cause, reachable over both the JSON and the protobuf representation (`-1` on the protobuf wire is `u64::MAX`).

New tests: `ra_tls::cert::{a_validity_bound_past_the_representable_range_is_an_error, a_validity_bound_inside_the_representable_range_is_accepted, signing_a_csr_with_an_unrepresentable_not_after_fails_without_aborting}` and `dstack_guest_agent::rpc_service::a_validity_bound_past_the_representable_range_is_rejected`.

`cargo test -p ra-tls -p dstack-guest-agent --all-features`: 38 + 147 passed.


kvinwang:
Regression sweep on this branch: the same 1016-vector adversarial request-contract matrix that found the abort now completes with **0 crashes and 0 server errors** across all 29 indexed guest-OS RPC methods (`Tappd`, `DstackGuest`, `Worker`, `GuestApi`), over both the JSON and the protobuf representation. Against `next` the same sweep aborts the process 4 times, all four on this root cause.

kvinwang:
### tdxlab hardware run: before and after

The new `tc-gos-contract-00{1..4}` acceptance cases (`test/core-coverage-expansion`) drive the full adversarial request-contract matrix against a lease-owned guest agent. Run on `tdxlab` (`phala-tdx-lab`), 2 workers, `local-simulator` fixture, prepared binaries built from the revision under test.

**Against the suite branch alone** (no product fix — run `cov841-4572fb4eb-gos`):

| case | status |
|---|---|
| tc-gos-contract-001 (Tappd) | **FAIL** — 2 violations |
| tc-gos-contract-002 (DstackGuest) | **FAIL** — 225 violations |
| tc-gos-contract-003 (Worker) | PASS |
| tc-gos-contract-004 (GuestApi) | PASS |

```
tc-gos-contract-002: {"detail": "transport failure: curl: (52) Empty reply from server",
                      "invariant": "L1", "label": "DstackGuest.GetTlsKey|not_after=u64-max|json"}
                     {"detail": "rejection body is 1048664 bytes",
                      "invariant": "L4", "label": "DstackGuest.GetTlsKey|alt_names=long-1m|json"}
```

The 225 is one abort plus the cascade: every later call in that sweep failed to connect, because the process was gone.

**With #1232 and #1234 merged in** (run `cov841v-1280b81da`, same host, same fixture):

```
PASS tc-gos-contract-001   The request-contract matrix held for every indexed method.
PASS tc-gos-contract-002   The request-contract matrix held for every indexed method.
PASS tc-gos-contract-003   The request-contract matrix held for every indexed method.
PASS tc-gos-contract-004   The request-contract matrix held for every indexed method.
PASS tc-gos-concurrency-001
PASS tc-gos-concurrency-003
SWEEP_DONE rc=0
```

PR #1234, opened 2026-09-20
fix(ra-rpc): bound the error text a failed call returns, and return it in the caller's representation

kvinwang:
Two defects on the shared pRPC error path, so they affect every dstack service — guest agent, gateway, KMS, VMM. Found by an adversarial request-contract sweep over the 110 RPC methods in `test-suites/catalog/api-inventory.json`.

## 1. A rejection echoes the whole value it rejected, into the reply and into the log

A deserializer names the value it refused, so the message is as long as the request field that produced it, and `encode_error` and the `rpc error:` log line both carry it verbatim.

```console
$ python3 -c 'import sys; sys.stdout.buffer.write(b"{\"path\":\"a\",\"subject\":\"b\",\"alt_names\":\"" + b"A"*(9<<20) + b"\"}")' > 9mib.json
$ curl -s -o /dev/null -w 'http=%{http_code} size=%{size_download}\n' \
    --unix-socket /var/run/dstack.sock -X POST -H 'Content-Type: application/json' \
    --data-binary @9mib.json http://localhost/prpc/Tappd.DeriveKey
http=400 size=9437272
$ awk '{ print length($0) }' simulator.log | sort -rn | head -1
9437466
```

One request: a 9 MiB reply and a 9 MiB log line. The default per-method limit is 10 MiB, so that is the ceiling. It is a response amplifier on every network-facing listener and a way to fill a service's disk — and on the guest agent it needs no credential at all, since `/var/run/dstack.sock` is bind-mounted into application containers by design.

Bounded at 2 KiB, dropping the **middle** rather than the tail. A deserializer puts the failure kind in front of the value and the input position behind it, so both ends carry information the caller needs:

```
invalid type: string "AAAA…... 1046445 bytes elided ...…AAAA", expected a sequence at line 1 column 70069
```

No real dstack error is near the bound, so nothing legitimate is truncated.

## 2. A JSON request could be refused in protobuf

`PrpcHandler::handle` decided the caller's representation with `request.json`, which is only true for a `GET` or an explicit `?json`. `handle_prpc_impl` decided the same question with the content type as well. Any failure raised *before* the dispatcher ran therefore took the first answer:

```console
$ curl -sD- -o- --unix-socket /var/run/dstack.sock -X POST \
    -H 'Content-Type: application/json' --data-binary @16mib.json \
    http://localhost/prpc/Tappd.DeriveKey
HTTP/1.1 413 Payload Too Large
content-type: application/octet-stream

\n&failed to read data: payload too large
```

A protobuf `ProtoError` under a content type the caller did not ask for, which no JSON client can parse. Every pre-dispatch failure is affected: `payload too large`, a certificate that will not parse, `failed to construct call`, and `invalid quote` — the one an attested caller most needs to be able to read.

Resolved once, before anything moves out of the request, and used on both paths.

## Verification

Three new tests in `ra-rpc/tests/status_codes.rs` and three unit tests for the bounding. Against `origin/next` the two behavioural tests fail:

```
test a_json_request_is_refused_in_json_before_the_dispatcher_runs ... FAILED
  assertion `left == right` failed: a JSON request must be refused with a JSON content type
    left: Some(ContentType(application/octet-stream))
   right: Some(ContentType(application/json))

test a_rejection_does_not_echo_the_whole_offending_value ... FAILED
  a rejection echoed 1048664 bytes
```

The unit tests cover the char-boundary case explicitly: the elided region is attacker-supplied, and byte-slicing a `String` panics when an index lands inside a character — which, with the workspace's `panic = "abort"`, would be a process abort rather than a failed request.

`cargo test -p ra-rpc -p dstack-guest-agent -p dstack-kms --all-features`: all green. `cargo clippy -p ra-rpc --all-features -- -D warnings --allow unused_variables`: clean.


kvinwang:
### tdxlab hardware run: before and after

The new `tc-gos-contract-00{1..4}` acceptance cases (`test/core-coverage-expansion`) drive the full adversarial request-contract matrix against a lease-owned guest agent. Run on `tdxlab` (`phala-tdx-lab`), 2 workers, `local-simulator` fixture, prepared binaries built from the revision under test.

**Against the suite branch alone** (no product fix — run `cov841-4572fb4eb-gos`):

| case | status |
|---|---|
| tc-gos-contract-001 (Tappd) | **FAIL** — 2 violations |
| tc-gos-contract-002 (DstackGuest) | **FAIL** — 225 violations |
| tc-gos-contract-003 (Worker) | PASS |
| tc-gos-contract-004 (GuestApi) | PASS |

```
tc-gos-contract-002: {"detail": "transport failure: curl: (52) Empty reply from server",
                      "invariant": "L1", "label": "DstackGuest.GetTlsKey|not_after=u64-max|json"}
                     {"detail": "rejection body is 1048664 bytes",
                      "invariant": "L4", "label": "DstackGuest.GetTlsKey|alt_names=long-1m|json"}
```

The 225 is one abort plus the cascade: every later call in that sweep failed to connect, because the process was gone.

**With #1232 and #1234 merged in** (run `cov841v-1280b81da`, same host, same fixture):

```
PASS tc-gos-contract-001   The request-contract matrix held for every indexed method.
PASS tc-gos-contract-002   The request-contract matrix held for every indexed method.
PASS tc-gos-contract-003   The request-contract matrix held for every indexed method.
PASS tc-gos-contract-004   The request-contract matrix held for every indexed method.
PASS tc-gos-concurrency-001
PASS tc-gos-concurrency-003
SWEEP_DONE rc=0
```

PR #1235, opened 2026-09-20
fix(gateway): a token that aborts the node, three admin RPCs that lie, and a renew lock that does not exclude

kvinwang:
Three independent defects found by a source audit of `dstack/gateway`, each with a test that fails against `next`.

## 1. A non-ASCII DNS token aborts the node, on every node, until it is deleted blind

`redact_token` byte-sliced a `String`:

```rust
format!("{}...{}", &token[..4], &token[len - 4..])
```

The token comes from `Admin.CreateDnsCredential` / `UpdateDnsCredential` with no ASCII validation on the way in. A byte index of 4 or `len - 4` landing inside a multi-byte character panics, and the workspace builds release binaries with `panic = "abort"`, so the gateway dies.

```
thread '...' panicked at gateway/src/admin_service.rs:938:34:
byte index 4 is not a char boundary; it is inside 'é' (bytes 3..5) of `abcé12345`
```

It is persistent and cluster-wide: the credential is stored and replicated *before* any read path formats it, so afterwards every `ListDnsCredentials` / `GetDnsCredential` aborts whichever node serves it. Recovery means calling `DeleteDnsCredential` — the one read-free path — blind. The trigger is a token pasted with a smart quote, a non-breaking space, or an emoji.

Fixed by counting and slicing `chars()`.

## 2. `ForceReleaseCertLock` reports a release it did not perform

`normalize_zt_domain` (lowercase, strip `*.`, strip the trailing dot) is applied by `AddZtDomain`, `GetZtDomain`, `UpdateZtDomain` and `DeleteZtDomain` — and was skipped by `RenewZtDomainCert`, `ForceReleaseCertLock` and `ListCertAttestations`. KV keys are built by raw interpolation.

`ForceReleaseCertLock` is the documented way out of a wedged renewal. Called with `APP.example.com` it deleted a key that never existed, returned `Ok(())`, and logged `Force released certificate lock for domain: …`. The operator is told the lock is gone while renewal stays wedged. `ListCertAttestations` silently reports a configured domain as never attested; `RenewZtDomainCert` at least errors.

Fixed by normalizing in all three. Force-release moved into a small free helper so the path is testable without standing up a `Proxy`, following the existing `merge_certbot_config` seam, and so the log names the key actually released.

## 3. The certificate renew lock is not mutual exclusion — including on one node

The read took a read guard, dropped it, then wrote under a write guard. With eight threads on one `KvStore`, **seven acquired the same lock**:

```
---- two_tasks_on_one_node_cannot_both_acquire ----
assertion `left == right` failed: round 0: exactly one task on a node may hold the lock
  left: 7
 right: 1
```

That is reachable in production: the background renew loop and an operator's `RenewZtDomainCert` run concurrently on one node. Two nodes both renewing is not benign — `authorize` calls `remove_txt_records`, which deletes *every* TXT record at `_acme-challenge.<domain>`, so each node destroys the other's in-flight challenge and both burn CA quota.

Two more holes in the same primitive:

- `get_cert_lock` used the lenient `decode`, so a corrupt record read as "unlocked" — fail-open on the one record whose job is to keep two renewals apart, for every node that replicated the torn value. Every other safety-critical read in that file deliberately uses `decode_strict`.
- `release_cert_lock` deleted unconditionally, so a renewal that outlived the timeout deleted the lock of the node that took over, letting a third renewal start alongside the second.

Fixed by holding one write guard across read, expiry check and write; reading with `decode_strict`; and having `release_cert_lock` take the value acquisition returned and delete only while that value is still visible, mirroring the existing `release_rotation_lock`. `Admin.ForceReleaseCertLock` keeps its unconditional delete under the explicit name `force_release_cert_lock`.

**On what this does and does not guarantee.** Cross-node exclusion is not achievable here, and the doc comment says so rather than claiming it. WaveKV is last-writer-wins with no compare-and-swap, so two nodes that have not yet exchanged the write both acquire; nothing short of consensus changes that. What the primitive now buys is a bound — concurrent renewal limited to the replication gap rather than the full renewal duration — and after-the-fact detection, because the loser's release finds someone else's lock and logs the takeover instead of clearing it. `two_unsynced_nodes_both_acquire_and_the_loser_detects_it_at_release` asserts that both nodes acquire: it documents the boundary, it is not a regression.

## Verification

Failing against `next`:

```
---- redacts_a_token_containing_multi_byte_characters ---- panicked: byte index 4 is not a char boundary
---- force_release_uses_the_key_the_lock_was_taken_under ---- panicked: the lock must be gone, not merely reported gone
---- force_release_refuses_a_malformed_domain ---- panicked: a malformed domain must be refused: "not a domain"
---- a_corrupt_lock_record_does_not_read_as_unlocked ---- panicked: a corrupt lock record must not hand out the lock
---- a_holder_that_was_taken_over_does_not_release_the_new_holder_lock ---- panicked: the new holder lock must survive the old holder release
---- two_tasks_on_one_node_cannot_both_acquire ---- left: 7, right: 1
```

`cargo test -p dstack-gateway`: 318 passed. `cargo clippy -p dstack-gateway -- -D warnings --allow unused_variables`: clean.

One judgement call worth flagging: `get_cert_lock` lost its last non-test caller and clippy's `dead_code` rejected it, so it is removed; tests assert through the live API or a local helper.


kvinwang:
Split into #1355, #1356 and https://github.com/Dstack-TEE/dstack/pull/1357 (stacked on #1356).

PR #1237, opened 2026-09-20
fix(kms): pin the onboarding key fetch, stop auto-bootstrap from rekeying, and guard the onboarding listener

kvinwang:
Four defects on the KMS onboarding and authorization paths, each with a test that fails against `next`.

## 1. Onboarding authorizes one connection and fetches the root private keys over another

`Keys::onboard` authorizes the source KMS on connection #1 — `RaClientConfig::builder().tls_no_check(true).cert_validator(...)` captures the peer's `VerifiedAttestation` and `ensure_kms_allowed` checks it. It then builds connection #2 with `RaClient::new_mtls` and issues `GetKmsKey` on *that* one — the RPC that returns the root CA private key and the root k256 private key.

`RaClient::new_mtls` sets `tls_no_check(true)`, `tls_built_in_root_certs(false)`, a client certificate and key, and **no `cert_validator` and no `tls_ca_cert`**. `try_validate_attestation` therefore returns at

```rust
let Some(validator) = &self.cert_validator else {
    return Ok(());
};
```

without looking at the peer at all, so `verify_server_attestation` (default `true`) is dead on that client. Nothing pins connection #2 to the peer authorized on connection #1 — not the certificate, not the attestation, not a key. `GetTempCaCert`'s material is used only to *mint the client certificate*; the KMS's own comment says that CA "authenticates nobody: it is handed to any caller".

Whoever answers the second connection chooses the replica's root keys. Every app that later asks that replica gets a disk key, an env X25519 key and a k256 key derived from them.

The prerequisite is only the ability to answer a TCP connection to `source_url` — which `docs/security/security-model.md` puts inside the threat model, and which `.agent/CODING_TASTE.md` reinforces ("endpoints are untrusted transport; the remote party's identity is verified cryptographically"). Combined with defect 3 below, no on-path position is even needed: an anonymous `Onboard.Onboard` pointed at a TCP proxy that forwards connection #1 to a genuinely allowlisted KMS passes `ensure_kms_allowed` verbatim, then answers connection #2 itself.

```
---- root_key_fetch_rejects_a_kms_it_did_not_authorize ----
called `Result::unwrap_err()` on an `Ok` value: KmsKeyResponse { temp_ca_key: "", keys: [] }
```
(the listener in that test presents a self-signed, unattested certificate)

Connection #1's validator now records the peer's attestation and SPKI; `ensure_kms_allowed` runs before anything is minted; and the key fetch goes through a client whose validator pins the peer's SPKI to the key that attestation was verified against. `RaClient::new_mtls` had exactly one caller and its contract was "authenticates the server not at all", so it is deleted rather than left as a trap.

## 2. Auto-bootstrap regenerates the root keys when any one of seven files is missing

`main.rs` gates onboarding on `!config.keys_exists()`, which requires **all seven** of `tmp-ca.crt`, `tmp-ca.key`, `root-ca.crt`, `root-ca.key`, `rpc.crt`, `rpc.key`, `root-k256.key`. With `auto_bootstrap_domain` set, `bootstrap_keys` then called `Keys::generate` and `keys.store(cfg)` unconditionally — while the *RPC* `bootstrap()` path guards with `bail!("KMS has already been bootstrapped")`.

Delete the least precious of the seven and restart, and the KMS silently writes a new `root-ca.key` and `root-k256.key`. Every app's disk key, env key and k256 key changes, existing encrypted disks become undecryptable, and the on-chain `kmsInfo.k256Pubkey` no longer matches. No attacker is required: a crash between `store_keys` and `store_certs`, or the disk rollback the security model explicitly grants the infrastructure operator, is enough.

```
---- auto_bootstrap_refuses_to_replace_existing_root_keys ----
assertion `left == right` failed
  left:  [… "…BG0wawIBAQQgEiF4Rsn0JneAbcpD…"]
  right: [… "…BG0wawIBAQQg+rHmTcF+EGT7Si2k…"]
```
— `root-ca.key` rewritten after deleting only `rpc.crt`.

Same guard as the RPC path.

## 3. The onboarding listener is unauthenticated plain HTTP on `0.0.0.0:8000`

`[core.onboard]` has `address = "0.0.0.0"`, `port = 8000` and no `tls` section; the `rpc` TLS figment and the `QuoteVerifier` are merged onto the *main* rocket only, so `handle_prpc_impl` always sees `attestation: None` and no `Onboard` handler calls `ensure_attested`. `dstack-app/docker-compose.yaml` publishes `8000:8000` — the same port guests later reach. So `Bootstrap`, `Onboard`, `GetAttestationInfo`, `Finish` and the plain `GET /finish` are callable by anyone who can route to it.

```
---- onboarding_listener_rejects_anonymous_callers_when_a_token_is_set ----
++     1: error sending request for url (https://attacker.example/prpc/GetTempCaCert?json)
++     3: dns error
>> outcome: success status: 400
assertion `left == right` failed
  left: Status { code: 200 }      # anonymous GET /
 right: Status { code: 401 }
```
The SSRF is live in that trace: the anonymous `Onboard.Onboard` really dialled out.

**On the fix shape.** Requiring attestation here is not a tightening, it is impossible: onboarding runs *before* the KMS has a certificate to serve, and the caller is a human operator with `curl` or the bundled `www/onboard.html`, not an attested peer. Binding to loopback is not viable either — the KMS runs in a container whose port the compose file publishes, so the tutorial's `curl` from the host would stop working. That leaves network placement plus an operator credential, and the project already has that mechanism: `dstack_api_auth`, used by the admin listener and by the VMM. It is now attached to every onboarding route, configured by `core.onboard.auth_token` / `htpasswd_file` / `DSTACK_KMS_ONBOARD_TOKEN`.

It **defaults to open**, not fail-closed, because every documented bring-up uses plain `curl` and backward compatibility is close to absolute here; an unset credential logs a startup warning naming exactly what is exposed. `onboarding_listener_stays_open_when_no_token_is_configured` pins that decision so it cannot be changed silently.

Correcting the severity: this is initialization hijack, denial of bring-up and SSRF — not key disclosure. `Bootstrap` still requires this KMS to be allowlisted by its own auth API and returns material that is published on-chain anyway, and with defect 1 fixed `Onboard` still requires an allowlisted attested source. The maintainers already treat anonymous reachability here as real: `0aabeecb39`, `865c50e3d3` and `d0456ef117` are all per-RPC mitigations for exactly this caller.

## 4. A UTF-8 boundary panic on every auth-API response

`upgrade_authority.rs`: `let short_body = &body[..body.len().min(512)];` byte-slices a `String`, on both the success and the failure branch, reached from the unauthenticated `GetMeta` and `GetAppEnvEncryptPubKey`. Release builds set `panic = "abort"`, so this kills the KMS.

```
panicked at kms/src/main_service/upgrade_authority.rs:110:27:
byte index 512 is not a char boundary; it is inside 'é' (bytes 511..513) of `{"status":"ok",…`
```

A localized proxy error page, an emoji in an `auth-simple` `reason` string, or a non-ASCII contract revert reason is enough.

## Verification

`cargo test -p dstack-kms`: 54 passed. `cargo clippy -p dstack-kms -- -D warnings --allow unused_variables`: clean. `cargo check --workspace`: clean.


kvinwang:
Split into focused PRs:

- #1305 — refuse auto-bootstrap over existing root keys
- #1306 — pin the onboarding key fetch to the authorized source KMS
- #1307 — guard the onboarding listener with an operator credential
- #1278 — truncate auth API bodies on a UTF-8 character boundary (already merged)

Closing this combined PR in favor of those changes.

PR #1246, opened 2026-09-20
test: cover the request contract, the KMS key hierarchy, and guest-agent concurrency

kvinwang:
Stacked on #841. Adds 19 cases in three new sections, and two product defects that the first of them found are already in review as #1232 and #1234.

## Why

Every RPC case in #841 names `catalog/api-inventory.json` as mandatory test data and states that the method is exercised with its request fields *"absent/default, valid, boundary-invalid and combined with an unknown field"* in both representations. The harnesses behind those cases send **one valid request and one wrong-typed value for the first request field**. So for all 110 indexed methods, the input classes where a service actually breaks — an out-of-range integer, an oversized string, a wrong protobuf wire type, a truncated body — are untested.

Two further gaps the chapters state but do not prove: six P0 KMS cases about live key hierarchies resolve to `cargo test -p dstack-kms`, whose own `remarks` field says *"no physical-origin trust claim is made"*; and the guest-OS chapter's only "concurrency" coverage is in-process Rust thread pools over parsers, with nothing driving the live agent under load.

### `01-guest-os/15-request-contract`, `03-kms/13-request-contract`, `04-gateway/11-request-contract` — 9 cases

One harness implementing the stated matrix, driven by the same inventory the per-method cases cite. Per field type: empty, wrong type, null, NUL byte, newline, path traversal, absolute path, control characters, 64 KiB, 1 MiB, odd-length hex, non-hex, 31/63/64/65-byte lengths, zero, negative, `u32::MAX`, `u64::MAX`, past `u64::MAX`, 4096-element repeated. Per representation: empty body, non-object, scalar, `null`, truncated object, trailing garbage, 2000-deep nesting, duplicate keys, 4 MiB body, truncated varint, runaway varint, overflowing length, length past the body, unknown wire type, group wire type, field number zero, field number past the declared range, 8 KiB of random bytes, JSON-as-protobuf, protobuf-as-JSON, absent content type, unrecognised content type.

It asserts six invariants that hold for every method whatever it means:

- **L1** a malformed request is never a 5xx, a dropped connection, or a transport failure;
- **L2** a rejected JSON request gets a structured JSON `error`, a rejected protobuf request gets a decodable `ProtoError`;
- **L3** the listener still answers a valid request after the whole matrix;
- **L4** a rejection neither echoes an unbounded amount of input nor *grows* with the field it refused;
- **L5** no call outlives its deadline;
- **L6** a rejection's `Content-Type` matches the representation the request arrived in.

L3 is a real assertion here: the workspace builds release binaries with `panic = "abort"`, so a reachable panic is a process abort.

`api-contract-matrix-policy.json` bounds how far each method may be driven. `full` is for a method whose effect stays inside the lease; `rejection-only` sends only what a service must refuse *before* it dispatches, so nothing can mutate; `skip` is for a method whose valid invocation ends the process or consumes an external resource, with the reason recorded per method. Every method in the inventory must carry a policy — one added without one fails the case rather than being silently skipped.

### `03-kms/14-key-hierarchy` — 6 cases

Against a live KMS: byte-identical derived material across a real restart (SIGTERM the recorded pid, re-exec the recorded config, PID proven changed); 9 app ids with all 36 pairs of environment public keys distinct and the issued `env_crypt_key` matching the published one for its own `app_id`; the `k256_signature` recovering to `GetMeta.k256_pubkey` over the documented preimage with six mutations rejected; `signature_v1` verifying, its timestamp strictly advancing, and a new signature failing under the old timestamp; 8 barrier-released concurrent `GetAppKey` calls hashing to one response.

**Four rows of the separation matrix are not covered and the case says so**: a second attested identity cannot be minted from the fixture. `app_id` and `instance_id` are not configuration — they are event-log payloads inside the captured `sdk/simulator/attestation.bin`, and `dstack-attest` enforces RTMR3 replay against the quote, so a second identity needs a second attestation fixture with an edited event log, recomputed RTMR3 and re-signed quote. That is provider work, written into the case's Postconditions with exactly what the fixture would need.

The harness's cryptography is anchored offline before it is trusted: 16 golden-vector rows including the committed `derive_k256_key` vector from `dstack/kms/src/crypto.rs`, cross-checked against a throwaway Rust k256/sha3 crate, and mutation-tested with six injected defects each caught by exactly the case that claims it.

### `01-guest-os/16-concurrency-and-robustness` — 4 cases

24 concurrent identical `GetKey`/`DeriveK256Key`/`DeriveKey` returning one key digest with path and `random_seed` controls; 24 concurrent `GetQuote` each binding its own `report_data` at bytes 568..632; 66 concurrent calls over 11 methods on all four listeners followed by a liveness probe and a `(pid, starttime)` identity check, so a restart cannot hide a wedge or an abort; a stalled log reader against an RSS bound and the per-chunk timeout.

`tc-gos-concurrency-002` (head-of-line blocking on the quote path) declares `hardware_required` and **refuses to PASS** when one quote costs less than 50 ms, reporting BLOCKED instead — against a simulator the quote is instant and the case would be meaningless. The case says why in its prose.

## Evidence

The assertions were shown to be real by breaking them on purpose: asserting `DeriveKey`'s certificate chain must also be identical → FAIL with *"1 distinct keys and 24 distinct certificate chains"*; shifting the report-data range by one byte → FAIL with *"24/24 concurrent quotes did not bind their own report data"*.

And on `tdxlab`, against unpatched product code:

```
[1/4] tc-gos-contract-001: FAIL    2 violations
[2/4] tc-gos-contract-003: PASS
[3/4] tc-gos-contract-002: FAIL    225 violations
[4/4] tc-gos-contract-004: PASS
```
```
{"invariant": "L1", "label": "DstackGuest.GetTlsKey|not_after=u64-max|json",
 "detail": "transport failure: curl: (52) Empty reply from server"}
{"invariant": "L4", "label": "DstackGuest.GetTlsKey|alt_names=long-1m|json",
 "detail": "rejection body is 1048664 bytes"}
```

The first is a process abort — `{"not_after": 18446744073709551615}` on the guest socket, which is bind-mounted into application containers by design. The 225 is that abort plus the cascade. Fixed in #1232 and #1234; with both merged, the same seven cases on the same host:

```
PASS tc-gos-contract-001 / -002 / -003 / -004
PASS tc-gos-concurrency-001 / -003
SWEEP_DONE rc=0
```

## One fix to #841 itself

`bun install --frozen-lockfile` fails on #841's current head, so `prepare-hardware-run.sh` cannot prepare a run from it. The three `package.json` files under `dstack/kms/auth-{eth-bun,simple,mock}` are byte-identical to `next`'s; the `bun.lock` files are not. One commit here takes `next`'s three lockfiles.

## Gates

`verify-registry --plan test-suites`: **0 findings**, 379 scripted / 379 promoted / 380 total, `reflects_committed_state: true`. `pytest test-suites/runner/tests -q`: 40 passed. `reuse lint`: compliant, 2528/2528. `ruff check` and `ruff format --check`: clean on the new harnesses.

## Not yet run on the lab, and the promotions say so

The KMS and gateway cases were validated against a mock KMS and offline golden vectors, not the lab fixture. Their `promoted-passing-cases.json` entries use `source_run: "kms-key-hierarchy-offline-mock-2026-09-19"` and their notes end with *"not yet executed on the lab fixture"* rather than claiming a run that did not happen. They should be re-promoted with a real `source_run` after the first lab sweep.

## Follow-up worth doing separately

`prepare-hardware-run.sh` keys the candidate guest image on the whole repo `HEAD`, so a commit that touches only `test-suites/` forces a full mkosi rebuild. The Rust side already keys its cache on a fingerprint of `dstack`, `Cargo.toml`, `Cargo.lock` and `rust-toolchain.toml` for exactly this reason; the image side should do the same over `os/` plus the build inputs.


kvinwang:
### tdxlab: 18/18 PASS

All the runnable new cases now execute on the `tdxlab` physical host against lease-owned fixtures. Run `cov841v-all-5860e60a4`, 1 worker, product binaries built from the revision under test with #1232 and #1234 merged in:

```
[1/18]  tc-gos-contract-001: PASS          [10/18] tc-kms-keyhier-001: PASS
[2/18]  tc-gos-contract-002: PASS          [11/18] tc-kms-keyhier-002: PASS
[3/18]  tc-gos-contract-003: PASS          [12/18] tc-kms-keyhier-003: PASS
[4/18]  tc-gos-contract-004: PASS          [13/18] tc-kms-keyhier-004: PASS
[5/18]  tc-gos-concurrency-001: PASS       [14/18] tc-kms-keyhier-005: PASS
[6/18]  tc-gos-concurrency-003: PASS       [15/18] tc-kms-keyhier-006: PASS
[7/18]  tc-gos-observability-slow-consumer: PASS   [16/18] tc-gw-contract-001: PASS
[8/18]  tc-kms-contract-001: PASS          [17/18] tc-gw-contract-002: PASS
[9/18]  tc-kms-contract-002: PASS          [18/18] tc-gw-contract-003: PASS
SWEEP_DONE rc=0
```

The six `tc-kms-keyhier-*` cases had only been validated against a mock KMS; this is their first run against a live one, and all eighteen promotions are re-recorded on this `source_run` with the offline-mock caveat removed.

`tc-gos-concurrency-002` is deliberately not in the set: it refuses to PASS where a quote costs less than 50 ms, so it needs a physical TDX guest rather than the simulator fixture.

Three harness bugs the lab found and fixed along the way, none of them product issues — the echo invariant was an absolute ceiling where it should have been a scaling comparison, `/proc/<pid>/fd` is owner-only and aborted a case the fixture otherwise passed, and the gateway and onboard fixtures publish their listeners with the `/prpc` mount already in the URL, so prefixing it again addressed routes rocket does not serve. That last one is worth noting as evidence the assertions are real: every one of the 626 gateway admin vectors came back as rocket's default HTML 404 page, and the matrix reported exactly that — an unstructured rejection under the wrong content type, and a listener that did not answer a valid request afterwards.

kvinwang:
### Regression: 65/65 PASS on the chapters that share fixtures with the new cases

The new cases are additive, but they lease the same `no-tee-dev` and `kms-ready` fixtures as the existing guest-OS and KMS chapters, so I swept every scripted non-hardware case on those two profiles on `tdxlab` (65 cases, excluding the 19 new ones):

```
by_status {'PASS': 61, 'ERROR': 4}
failed ['tc-kms-auth-005', 'tc-kms-auth-006', 'tc-kms-auth-009', 'tc-kms-runtime-003']
```

The four errors are environmental, not a regression:

```
RuntimeError: Foundry contract submodules are not initialized:
  lib/forge-std/src/Test.sol,
  lib/openzeppelin-contracts-upgradeable/contracts/proxy/utils/UUPSUpgradeable.sol, ...
```

After `git submodule update --init --recursive` in `dstack/kms/auth-eth` and `bun install --frozen-lockfile` in the three Bun packages, all four pass:

```
[1/4] tc-kms-auth-005: PASS
[2/4] tc-kms-auth-006: PASS
[3/4] tc-kms-auth-009: PASS
[4/4] tc-kms-runtime-003: PASS
SWEEP_DONE rc=0
```

**65/65.**

Worth noting for #841 independently of this PR: `prepare-hardware-run.sh` provisions the Foundry submodules and the Bun packages, and `prepare-run.sh` does not — so the contract and auth-backend cases error on a plain preparation rather than reporting a named prerequisite. Either `prepare-run.sh` should provision them too, or those cases should report BLOCKED with the reason instead of a `RuntimeError`.

kvinwang:
### The hardware case found a real product defect, and then named it

`tc-gos-concurrency-002` is the one case here that needs a physical TDX guest. Its first run on `tdxlab` failed with

```
15/43 concurrent DstackGuest.GetQuote under load calls failed:
  RemoteDisconnected: Remote end closed connection without response
```

and the case left it there. Every reading of that sentence is about the transport, and the fixture publishes a QEMU slirp forwarded host port into the guest, so the obvious suspect was the forwarder rather than the agent.

It was the agent. Driving the same 24-way load **four ways against one lease-owned guest** — over the forwarded port, over `/var/run/dstack.sock` from inside the guest, over the agent's own TCP listener, and through the fixture's socket bridge — drops **15 to 21 of 43 every time**, and the guest's journal says why:

```
dstack-guest-agent.service: Watchdog timeout (limit 30s)!
dstack-guest-agent.service: Killing process 533 with signal SIGABRT
```

The systemd heartbeat is an HTTP `Worker.Version` request the agent sends to its own external listener, so a runtime parked on the global quote mutex cannot answer it. Control: 24-way `Version` under the same conditions answers **579 405 calls at a 2 ms p95** — so it is the blocking quote, not the concurrency.

Product fix: #1256. With it, all 43 are answered, `Version` holds at 2 ms p95, and four consecutive loads leave the agent in the process it started as.

Two corrections to the case itself, now committed here:

- It reads the agent's PID and start time from inside the guest over the fixture's `ssh_argv`, before and after the load, and checks that **first** — a dropped connection reads as a transport fault until you know the agent it was talking to was killed. Best-effort, so a fixture that publishes no guest access records nothing rather than blocking a case whose assertions are all measured over RPC.
- The worker count it reasons from was wrong. `#[rocket::main]` builds the runtime before `load_config_figment` runs, so the `workers = 8` in `dstack.toml` never reaches it: the 2-vCPU guest was measured with **two** `rocket-worker` threads. The 24-thread load saturates either way, but the derivation should say the true one.

kvinwang:
### All 19 cases now pass on tdxlab, including the hardware-only one

`tc-gos-concurrency-002` was the last case without a green run — it refuses to PASS where a quote costs less than 50 ms, so it needs a physical TDX guest, and against an image built from `next` it correctly fails by naming the agent restart it found.

Against a guest image built by mkosi from this branch plus #1256:

```
[1/1] tc-gos-concurrency-002: PASS
HWDONE rc=0
```

> A trivial method stayed within its latency bound while the attestation path was saturated, in the agent process the case started with.

So the full set is **19/19 on tdxlab**: 18 on lease-owned simulator and isolated-component fixtures (run `cov841v-all-5860e60a4`), and this one on a physical TDX guest (run `cov841v-hw-ca26fb35e`).

PR #1249, opened 2026-09-20
test: property-test every parser that consumes attacker bytes

kvinwang:
The repository has no fuzzing and no property testing of any kind — `rg 'fuzz_target|cargo-fuzz|proptest|quickcheck|arbitrary|afl'` over the whole tree returns nothing, and there is no `fuzz/`, `corpus/` or `testdata/` directory. Every parser in the attestation and measurement path consumes fully attacker-controlled bytes, and release binaries build with `panic = "abort"`, so a panic in one of them is a `dstack-verifier` or `dstack-kms` process abort rather than a failed request.

**This branch is stacked on #1231 and #1236, and contains #1248.** Four of the corpus patterns are exactly what those PRs fix, so on bare `next` the suite aborts or hangs. It shrinks as they land.

## Design

`proptest 1.11` as a workspace dev-dependency (`default-features = false, features = ["std", "bit-set"]` — no `fork`/`timeout`, which would fork a process per case). Chosen over a hand-rolled `arbitrary` driver for shrinking: every counterexample below is a *minimised* one.

A small dev-only crate `dstack/crates/prop-harness` (173 lines) owns the shared mechanics: fixed seeds, bounded case counts, and `within_deadline`, which runs the parser on a worker thread so a **non-terminating** parser fails instead of stalling CI. Failure persistence is off, so CI never wants to write `.proptest-regressions`.

The property modules live per parser *source file* (`dstack-mr/src/{tdvf,kernel,tdx,sev}/proptests.rs`, `tpm-qvl/src/verify/proptests.rs`) rather than per crate. Most targets — `parse_kernel_cert_table`, `parse_tpm_attest`, `authenticode_sha384_hash`, the `ovmf_footer_*` set — are module-private, and a child `#[cfg(test)] mod proptests;` reaches them with **zero production visibility changes**. The alternative was making parsers `pub(crate)`/`pub` for tests.

**No `cargo-fuzz` directory**, for the same reason: reaching those private parsers from a separate fuzz crate means making them public — permanent API surface for an opt-in nightly job. Instead `DSTACK_PROPTEST_CASES` overrides the case count and scales the deadline budget with it:

```
DSTACK_PROPTEST_CASES=1000000 cargo test -p dstack-mr proptests
```

Same code, same strategies, deeper search, no new surface.

## The strategies had to be structure-aware, and I measured that

The first cut drew all six `TdvfSection` fields independently. Instrumented: of 256 generated images, 16 parsed and **0 sections ever reached the hashing loop** — the search was only exercising rejection paths. The checks are conjunctive, so a uniform section passes about 1 in 12 and a 3-section image about 1 in 2000. Rebuilt as *valid section plus one mutated field*: 72/256 parse, 311 sections walked. Same treatment for `cc-eventlog` (32 → 89 decodes).

## Mutation-tested, so "all green" means something

Each in-flight fix was reverted and the property confirmed to go red:

| reverted | caught by |
|---|---|
| zero-length GUID entry (#1236) | `the_guid_table_walk_always_terminates` + corpus |
| `MAX_MEASURED_PAGES` (#1236) | `the_section_table_walk_never_panics` (30 s deadline) |
| TD HOB range cap (#1236) | `the_witness_decoder_terminates_on_arbitrary_varints` + corpus |
| PE header bounds (#1231) | `authenticode_never_panics_on_arbitrary_pe_headers` + corpus |
| `sha256sum` fix (#1248) | `a_manifest_naming_one_file_twice_is_rejected` |
| cert-table cap (#1248) | `rejects_a_certificate_table_that_decodes_to_more_than_it_contains` |

## What it found

The two product defects are in #1248. Probed and found **clean**, now pinned as regression tests: CBOR nesting 1k–400k deep → ciborium `RecursionLimitExceeded`; a declared byte-string length of `u64::MAX` → EOF rather than a preallocation; `parse_tpm_attest` PCR-index amplification bounded at ~32×.

### The V1 runtime-event digest preimage is not injective — asserted, not fixed

`cc-eventlog`'s V1 preimage is `type.to_le_bytes() || b':' || name || b':' || payload`, an unlength-prefixed concatenation. Run as a property over `(name, payload)` pairs drawn from `{'a', ':'}`, proptest minimises to:

```
("::", b":")  vs  (":::", b"")     → both encode to TYPE + five colons
```

Not hypothetical: `emit_event` lets a workload choose the name, and `key-provider` payloads are JSON containing `:`. The impact is bounded — RTMR3 replays over *digests*, so two logs replay identically while displaying different `(name, payload)` to a relying party — and the encoding is frozen, so this is encoded as `the_v1_digest_preimage_is_not_injective`, asserting the collision holds, with both the minimal case and a realistic `key-provider` re-split. `the_v2_digest_preimage_is_injective` runs the same search over `canonical_event_json_v2` widened with JSON metacharacters, and holds.

## Corpus

17 files under `dstack-mr/tests/corpus/{tdvf,pe,td_hob_witness,ovmf}/` with a `README.md` documenting each pattern's construction, read by table-driven `*_corpus_never_panics` tests. All the known patterns are present — zero-length GUID entry, section past the blob, `memory_data_size = 0x1000_0000_0000`, `memory_address = 0xffff_ffff_ffff_f000`, a 50 000-range TD HOB witness, PE `SizeOfHeaders` of `0xFFFF_FFF0` and of `0`, a file of length exactly `lfanew + 88` — plus `guid_table_entry_len_17`, `data_offset_size_overflow`, `page_arithmetic_overflow`, `section_past_image` and four OVMF ones. `ranges_50000.bin` is 191 KB because a truncated copy does not reproduce: the decode has to actually reach the ranges.

## Runtime

**0.83 s** of test execution for all 45 property tests; **3.65 s** wall including cargo overhead. Target was well under a minute.

Soak at `DSTACK_PROPTEST_CASES=300000` (~1170× the gate): all six crates clean, zero failures. One intermediate 200k run flagged a false positive in a reference model I had written for `sha256sum` — it was line-agnostic while the fixed parser is correctly line-based — so the model was replaced with the direct invariant rather than letting a near-copy of the implementation check the implementation.

## Verification

```
cargo test -p dstack-mr -p cc-eventlog -p dstack-attest -p dstack-types -p sev-snp-qvl -p tpm-qvl
  → 249 passed, 0 failed (1 ignored: the pre-existing network-downloading tdvf_parse.rs)
cargo clippy --all-features -- -D warnings --allow unused_variables
  → exit 0
```
Also clean under CI's stricter `-D clippy::expect_used -D clippy::unwrap_used` for every crate touched.

## Two notes for review

The `dstack-attest` round-trip is stated as `to_bytes ∘ from_bytes ∘ to_bytes == to_bytes` rather than value equality: `Attestation` has no `PartialEq` and SCALE deliberately drops `#[codec(skip)]` fields, so value equality is the wrong property. Encoding stability is the one the wire format actually needs, since a re-encode that moved would change the CSR bytes a KMS signs.

If #1231 or #1236 land in a different shape than they have now, the four `dstack-mr` corpus assertions that check for specific error substrings will need re-checking.


kvinwang:
Closing: the defects this suite found now have focused regression tests in their own fix PRs, #1236 has merged in a different shape, and the branch conflicts with next. A smaller follow-up covering only the pre-signature parsers can be opened if we still want property tests.

PR #1253, opened 2026-09-20
fix(gateway): every connection forks a process when wg show fails, and AcmeInfo serves unbounded history to anyone

kvinwang:
Three availability defects on the gateway's routing and state paths, plus one cluster-trust finding I could confirm but not fix here.

## 1. When `wg show` fails, every inbound connection forks a process — under the global routing mutex

`select_top_n_hosts` is called with the `ProxyState` mutex held, and it reaches `LatestHandshakesCache::latest` → `fetch_latest_handshake_timestamps` → `cmd_lib::run_fun!(wg show …)`, a blocking fork/exec/wait on a tokio worker.

On the "can the cache really stay empty forever" question — **it can, and the refresh task does not save you.** `TtlCell::get_allow_stale` returns `Empty` until `set()` is called, and `set()` is only reached through `refresh_blocking` *after* the producer succeeds. Both producers are the same `wg show`: the eager preload in `ProxyInner::new` and the periodic refresh task. If `wg show` can never succeed — wrong `core.wg.interface`, no WireGuard module, `wg` not on `PATH`, a container without `CAP_NET_ADMIN` — nothing ever calls `set()`, the cell is `Empty` for the life of the process, and **every** inbound connection forks. Not "one connection in a starved window".

And on a healthy host the preload fills the cell before the proxy listener starts, so the synchronous fallback was buying nothing there either.

The codebase already knew: `main_service.rs` and `proxy/health_check.rs` both say *"on a cold cache it shells out to `wg show` synchronously — neither belongs inside the mutex the routing path takes once per connection."* The routing path was the one place that did it.

```
200 cold reads took 182.864481ms: the routing path is spawning a process per connection
```
plus 200 `Unable to access interface: Operation not permitted` lines from the forked `wg` processes. After the fix, 500 iterations finish in 0.00s.

`latest` is now cache-only — fresh, then stale, never a producer call. Serving the *stale* value rather than empty is deliberate: every consumer thresholds on handshake *age*, so a stale map ages the fleet uniformly, while empty would drain routing on the first failed refresh.

## 2. `Admin.Status` writes to the KV store with the routing lock held

`refresh_state` was called under `self.state.lock()`, and it performs one ephemeral KV write per instance. After defect 1 the cost is no longer the fork but **lock composition**: each `sync_instance_handshake` takes `self.ephemeral.write()`, the same lock a sync round holds while merging a multi-megabyte inbound envelope. Holding `ProxyState` across N acquisitions of it lets one peer's merge stall every proxied connection. `route_index.rs` reaches the same `status()`, so a dashboard page load does it too.

```
panicked at gateway/src/main_service/tests.rs:2715:
the routing lock is held while the refresh waits on the KV store
```
(the test holds the ephemeral store's write lock, spawns the refresh, asserts it really is parked, then asserts `try_lock()` on the routing mutex still succeeds)

`refresh_state` moved from `ProxyState` to `ProxyInner`: it reads the handshake cache, takes the routing lock only to copy the instance-id/timestamp pairs, drops the guard, then writes.

**One edit outside my scope, unavoidably:** deleting `ProxyState::refresh_state` makes `admin_service.rs:54-56` not compile, so two lines there had to move. #1235's `admin_service.rs` hunks all start at line 647 or later, so there is no textual overlap. **Left for #1235's owner:** `get_info` and `get_meta` still take the routing lock and then read handshakes under it — harmless now that the read cannot fork, but they should be reordered the same way.

## 3. `AcmeInfo` is unauthenticated and serves the whole attestation history

`save_cert_attestation` writes one key per renewal and never prunes. `acme_info` iterated all of it for every configured domain, cloning `public_key` + `quote` + `attestation` into one response, and unlike `get_peers` it does not call `ensure_from_gateway`. It is reachable at `/.dstack/acme-info` on the public proxy. `Admin.ListCertAttestations` has a `limit`; this had none.

```
an unauthenticated request drew 10000 history entries
```
— a ~20 MB response per unauthenticated request.

One nuance worth stating plainly: growth is *renewal*-driven, not attacker-driven. A stranger cannot add entries, only force the walk and serialization of what renewals have accumulated.

Capped at 32 per domain, newest first — per-domain rather than whole-response, so adding a domain cannot silently stop older ones being served.

**The endpoint stays unauthenticated, deliberately.** `ct_monitor` fetches `/.dstack/acme-info` over the public proxy to learn which keys may legitimately appear in a CT log; it runs outside the cluster and holds no gateway certificate. `docs/deployment.md` and `certbot/cli/README.md` document it as the public interface. Removing access would break CT monitoring, and 32 entries is far more than that consumer needs for a 90-day certificate.

**Residual, belongs in #1235's `kv/mod.rs`:** the cap bounds the response but not the walk — `list_cert_attestations` still iterates and msgpack-decodes every history key under the persistent read lock, measured at 24.6 ms (release) for 10 000 entries. The clean fix is a `limit` parameter at `kv/mod.rs:2248`; #1235's hunks are at 2118/2409/2494, so that line is clear if its owner wants it.

## 4. `sender_id` in the WaveKV sync routes: (a) by design, (b) real and not fixable here

I wrote both demonstration tests and **both passed against unfixed code**, so both behaviours are real. I removed them rather than commit tests asserting an insecure behaviour.

**(a) "any same-app-id certificate holder can merge into the store" — by design, and the proposed fix would deadlock the cluster.** The module header's premise holds: a same-app-id certificate is issued to a CVM running the same measured code, dstack derives app keys per app-id, and such a node is entitled to join. More decisively, requiring `sender_id ∈ peer_ids()` would break cluster growth: `add_peer` is reached only from `register_peer_url`, `sync_node` does not add to the peer set, and nothing on the inbound path does either — so a bootnode only learns a joiner through the very envelope the check would reject.

**(b) `sender_id` is unbound to the certificate — CONFIRMED.** `wavekv` records `peer_acks` keyed by `env.sender_id`, and the tombstone-GC watermark is a `min` over those. Two things an honest-id peer cannot do:

- **The removal lockout is bypassed by changing one integer.** `refuse_removed_sender` keys on the self-asserted id, so a removed node — the one adversary the project built an explicit mechanism for — re-enters under any live node's id. It already holds every uuid, because uuids are replicated in `node/{id}` records.
- **Forging a member's id raises that member's contribution to the GC minimum.** A peer using its own id can only raise its own term; the victim's real report still pins the min. Spoofing removes that pin and collects tombstones a live node has not seen — the resurrection the ack gate exists to prevent.

**I did not build the binding, because it cannot be built in this file.** A gateway peer certificate carries app-id (per *app*, not per node) and app-info. The only per-node value is `instance_id`, and the gateway never learns its own (`main.rs` takes only `.app_id` from the guest agent), and nothing records a node-id → instance-id mapping to compare against. A sound fix spans `main.rs`, `main_service.rs` and `kv/mod.rs` (#1235's file), with a legacy-empty compat case that keeps the hole open through a rolling upgrade. That is a design change, not a patch, and I did not want to ship a TOFU-on-SPKI approximation that breaks on certificate rotation.

I also evaluated and rejected forcing `push_only` for non-member senders: merging alone does not bump acks, so a bootnode would never advance `acks[joiner]` and the joiner's tombstones could never be collected.

**What is shipped** is the one guard that is cheap, clearly correct and needs no new identity: refuse an envelope claiming **this node's own id**, alongside the existing id-0 check. Nothing syncs to itself, so such an envelope is either a misconfigured peer or a caller reaching for the ack bookkeeping that feeds every peer's watermark. Pre-fix: `left: Status { code: 200 }, right: Status { code: 400 }`.

## Verification

`cargo test -p dstack-gateway`: 313 passed. `cargo clippy -p dstack-gateway -- -D warnings --allow unused_variables`: clean.

**One conflict to expect with #1239:** it inserts `enum WgApply` immediately before `const HANDSHAKE_CACHE_TTL` in `main_service.rs`, and this inserts `MAX_ACME_HIST_KEYS_PER_DOMAIN` at the same point. Keep both. No other overlap — #1239's `main_service.rs` hunks are at 96/485/722/2304/2329 and mine at ~305, ~660, ~965 and ~2517; nothing under `proxy/` (#1245) is touched.


kvinwang:
Follow-up PR #1274 completes the remaining Admin.Status work: KV handshake reads and node enumeration are outside ProxyState, and status runs off the async worker.

PR #1256, opened 2026-09-20
fix(guest-agent): a container can make systemd SIGABRT the guest agent by asking for quotes

kvinwang:
Found by running the new `tc-gos-concurrency-002` acceptance case on real Intel TDX hardware, and then chasing down what the drops actually were.

## What happens

`tdx_attest::get_quote` takes a **process-global `std::sync::Mutex`** and then blocks — on configfs/TSM, or on a vsock connection to the host's QGS with no connect, read or write timeout. Exactly one caller treated it as blocking work: `Attest` on the v1 surface, whose comment names the problem precisely:

> takes a global mutex and then blocks in an ioctl. On the async executor that parks a worker thread for the duration and stalls every other connection this agent is serving.

`GetQuote`, the v0 `Attest`, `Tappd.TdxQuote`, `Tappd.RawQuote`, `Worker.GetAttestationForAppKey`, `issue_cert` and `get_info` all ran it straight on the async executor.

**And the runtime is narrower than it looks.** `#[rocket::main]` builds the runtime *before* the configuration is read, so the `workers = 8` in `dstack.toml` never reaches it — a CVM gets one worker per vCPU. A 2-vCPU guest was measured running two `rocket-worker` threads.

## Measured on TDX hardware

One quote costs **1.02 s** on a 2-vCPU guest. 24 concurrent `GetQuote` calls over the bind-mounted `/var/run/dstack.sock`:

- **20 of 43** calls dropped;
- `Version` went from 0.4 ms to seconds.

And the stall reaches the agent's own watchdog. The systemd heartbeat is an HTTP `Worker.Version` request the agent sends to **its own external listener**, so a runtime parked on the quote mutex cannot answer it:

```
dstack-guest-agent.service: Watchdog timeout (limit 30s)!
dstack-guest-agent.service: Killing process 533 with signal SIGABRT
```

systemd kills the agent and restarts it, and every open connection is reset. **Any container in the CVM can do this** — the socket is bind-mounted into application containers by design — and the same load over the external TCP listener does the same thing.

### Ruling out the transport

The first observation came through a QEMU slirp forwarded host port, which has its own limits, so the same 24-way load was driven **four ways against one lease-owned guest**: over the forwarded port, over `/var/run/dstack.sock` from inside the guest, over the agent's own TCP listener, and through the fixture's socket bridge. **15 to 21 of 43 dropped every time.**

Control: 24-way `Version` under the same conditions answered **579 405 calls at a 2 ms p95**. So it is the blocking quote, not the concurrency, and not the transport.

## The fix

Put the hop where the blocking is, not at each caller: `quote_response` and `attest_cvm` take `spawn_blocking` themselves, and `issue_cert` and `get_info` wrap their platform calls, so a new caller cannot forget.

`identity()` deliberately stays on the executor — as its own comment says, the retry throttle caps it at one quote per interval for the whole process.

## After

Same load, same guest: **all 43 calls answered**, the `Version` probe stays at **2 ms p95** against a 1.0 s bound, and four consecutive 24-way loads leave the agent **in the process it started as**.

`cargo test -p dstack-guest-agent`: 147 passed. `cargo clippy -p dstack-guest-agent -- -D warnings --allow unused_variables`: clean.

## Worth a separate look

`#[rocket::main]` building the runtime before `load_config_figment` means `[core] workers` in `dstack.toml` is silently inert. This PR does not change that — it removes the reason it mattered — but the configuration key is currently a lie, and either it should take effect or it should go.


kvinwang:
### Full hardware measurement, four transports, before and after

Lab `tdxlab`, one lease (`lease-tc-gos-concurrency-002-3cee0914b02f`), image `dstack-mkosi-4572fb4eb`, 2-vCPU TDX guest. Load: a static Go binary scp'd into the guest, one connection per call, 24 threads × 20 s, with `Version` probed every 100 ms on its own connection. The guest has `python3` and `curl`; the Go tool was used for precise concurrency and error classification.

**Candidate agent, four transports — the drops are not the transport:**

| path | load ok/total | probe p50 / p95 | agent identity |
|---|---|---|---|
| host → forwarded port 19003 (slirp) | 27/44 (**17 failed**) | 7.20 s / 18.45 s | 533→1692 **restarted** |
| **in-guest → `/var/run/dstack.sock`** | 23/43 (**20 failed**) | 1.05 s / 22.36 s | 1692→1939 **restarted** |
| in-guest → agent TCP `:8090` | 28/43 (**15 failed**) | — / 25.62 s | 2379→2835 **restarted** |
| in-guest → fixture bridge `:3000` | 22/43 (**21 failed**) | — / 19.47 s | 2835→3207 **restarted** |
| in-guest unix, `origin/next` build swapped in | 28/43 (**15 failed**) | — / 22.57 s | 4365→4604 **restarted** |

Unloaded `Version` p95 is 0.8–1.3 ms on every path. One `GetQuote` costs **1.02 s**.

**Controls, same guest, same session:**

| load | result | identity |
|---|---|---|
| 24-way `Version`, no quote, 20 s | **579 405 / 579 405 ok**, probe p95 **2 ms** | unchanged |
| 2-way `GetQuote`, 20 s | 21/21 ok, but probe p50 already **0.92 s** | unchanged |

The first rules out concurrency. The second shows two concurrent quotes already consume the whole runtime.

**With this branch, same guest, same four transports:**

| path | load ok/total | probe ok/total | probe p95 | identity |
|---|---|---|---|---|
| forwarded port 19003 | **43/43** | 185/185 | 4 ms | 4736 |
| in-guest unix socket | **43/43** | 186/186 | 2 ms | 4736 |
| in-guest TCP `:8090` | **43/43** | 186/186 | 2 ms | 4736 |
| in-guest bridge `:3000` | **43/43** | 185/185 | 2 ms | 4736 |

Same pid **and** start-time tick (4736 / 109370) across all four consecutive loads, `NRestarts=0`, threads 3→13 (the blocking pool), RSS +6 MB.

Also run through the real acceptance harness against the leased guest: **PASS** with the fixed binary swapped in via a `/run/systemd/system` drop-in (`agent_restarted: false`, loaded p95 3.8 ms), **FAIL** with the in-image binary, naming the restart.

A regression test, `a_blocking_quote_does_not_park_the_runtime`, runs on a `current_thread` runtime and was verified to fail ("the executor ran 0 times") when the hop is removed.

kvinwang:
### End-to-end on a guest image built from this branch

The measurements above swapped the fixed binary into a running guest with a systemd drop-in. This is the same case, through the acceptance runner, against a **guest image built by mkosi from this branch's source** — so the agent under test is the shipped one, not a hot-patched one:

```
test-suites/runner/dstack-test sweep --plan test-suites --run-id cov841v-hw-ca26fb35e \
  --workers 1 --runtime-manifest ~/cov841v-hw-ca26fb35e-runtime.json \
  --case tc-gos-concurrency-002

[1/1] tc-gos-concurrency-002: PASS
HWDONE ca26fb35e rc=0
```

> A trivial method stayed within its latency bound while the attestation path was saturated, in the agent process the case started with.

Against an image built from `next` the same case fails, naming the restart:

> the agent did not survive the load: `{'pid': 531, 'starttime_ticks': 2285}` became `{'pid': 1455, 'restarts': 1, 'starttime_ticks': 12937}`. 10 of 43 saturating calls and 1 of 1 probes were dropped with it. Agent journal: `… Killing process 531 with signal SIGABRT … Failed with result 'watchdog'`

kvinwang:
Follow-up commit a193b5eb70 fixes the residual startup-identity failure path: later anonymous Info retries are single-flight and run on spawn_blocking; cancellation retains the guard.

PR #1257, opened 2026-09-20
fix(os): stop zapping an unencrypted data disk on upgrade

kvinwang:
With `dstack.storage_encrypted=0`, `mount_data_disk` formats the raw data disk, so it has neither a LUKS header nor a `dstack-data` partition. On the next boot after an image upgrade, `choose_data_device` treats it as:

- ext4: an empty disk, and runs `sgdisk -Z` on it (data loss);
- zfs: a foreign partitioned disk (zpool writes its own GPT), and fails the boot.

Use the whole disk when `blkid -p` finds a filesystem, or when a partition is a `zfs_member`. Only blkid's "found nothing" (exit 2) counts as empty, so an uncertain probe never leads to a zap.

Tested on loop devices (ext4 whole disk, zpool-shaped GPT with ZFS labels, blank disk, foreign GPT): both upgrade cases fail on `next` and pass here; blank and foreign disks behave as before.

The other three fixes originally in this PR are split out: https://github.com/Dstack-TEE/dstack/pull/1320, https://github.com/Dstack-TEE/dstack/pull/1321, https://github.com/Dstack-TEE/dstack/pull/1322.


kvinwang:
Closing: `storage_encrypted=0` is only used for development, and production disks always carry a LUKS header, which step 3.1 already handles. Not worth the extra detection logic.

PR #1263, opened 2026-09-20
fix(guest-agent): an anonymous Info drives a quote and a KMS round trip, and the supervisor fsyncs its log every 8 KiB

kvinwang:
Five audit findings worked to a conclusion: **three confirmed and fixed, two refuted with the evidence.** The refutations land as a pinning test and a doc comment rather than nothing.

`cargo test -p dstack-guest-agent -p supervisor`: 153 + 4 passed. `cargo clippy -p dstack-guest-agent -p supervisor -- -D warnings --allow unused_variables`: clean.

### 1. Each anonymous `Info` could spawn a fresh quote plus a KMS round trip

`maybe_request_demo_cert` guarded only on "is `demo_cert` empty" — no in-flight guard, no failure backoff. The fan-out per spawn is `request_demo_cert` → `issue_cert` → `platform.certificate_attestation()` (a hardware quote under the process-global `QUOTE_LOCK`) **plus** `cert_client.sign_csr()` (an RA-TLS round trip to the KMS). It is reached from `get_info`, i.e. every anonymous `Worker.Info`, every dashboard `GET /`, and every vsock `GuestApi.Info`.

So while the KMS is unreachable — exactly when the guard matters — each request adds another.

```
concurrent_info_calls_request_at_most_one_demo_cert
  4 certificate attestations in flight at once     (capped only by the worker count)
the_demo_cert_..._throttle_window
  9 requests from 9 calls
```
Both are 1 after.

Fixed with a single-flight `in_flight` flag **plus** a `DEMO_CERT_RETRY_INTERVAL` (30 s) failure throttle — the same shape and floor as the existing `IDENTITY_RETRY_INTERVAL`. No lock is held across the request and the two are never held at once.

### 2. The identity cache existed for anonymous `Info` and anonymous `Info` did not use it

`get_info` called `info_attestation()` unconditionally, and `RealPlatform::attestation_for_info` is `Attestation::local()` = `Attestation::quote(&[0u8;64])` — a real quote under `QUOTE_LOCK`. `get_info(external=true)` is reached by `Worker.Info`, the dashboard and vsock `GuestApi.Info`. Only v1's `info_response` used the cache the comments describe.

```
an_external_info_answers_from_the_identity_cache
  8 attestations from 8 anonymous calls
```
1 after (the boot decode).

Split on what the call actually serves: `tcb_info` is the only field that needs an attestation, and `external` already blanks it unless `public_tcbinfo` — so the hiding path now answers from `identity()` and never touches the platform.

**The frozen-API check that shaped this:** the *serving* path still attests and decodes identity from that attestation rather than from the cache. Otherwise a surface that always attested would start returning the cache's throttled "identity unavailable" after one bad decode. `the_cached_info_reports_what_the_attesting_path_reports` pins field-for-field equality between the two.

### 3. The supervisor fsync'd and reopened its log roughly every 8 KiB of child output

`try_redirect` treated `is_modify()` on the log path as "logrotate happened", and `notify` 8's inotify backend emits `Modify(Data(Any))` for the supervisor's **own** `write_all`. Measured with `strace -f -e trace=fsync,openat` on the real binary, 8 MiB through one supervised child, nothing rotating:

| | fsync | opens of `chatty.log` |
|---|---|---|
| before | **1145** | 1146 |
| after | **0** | 1 |

(log size identical, 8 200 192 B). Both calls are blocking `std::fs` on a tokio worker, and there are two such loops per managed process.

**The obvious fix would have broken logrotate.** Probing `notify` 8.2: a rename emits `Modify(Name(From))`, `Modify(Name(To))`, `Modify(Name(Both))` — **not** `Remove` — so matching `is_remove()` alone would have missed `logrotate(8)`'s default mode entirely. The predicate is now `Remove(_) | Modify(Name(_))`.

A truncate-in-place rotation needs no reopen at all, because the fd is `O_APPEND` — which is what `dstack-vmm`'s own `logrotate` module and `logrotate(8) copytruncate` do, and what `vmm/src/logrotate.rs`'s module doc already states. The stale "and reopens them when they change" half of `vmm/src/app.rs:1891` is corrected here.

Three of the four new tests fail on the old predicate, including `streaming_output_does_not_reopen_the_log` (56 reopens for 1 MiB in the harness).

### The `report_data` length divergence is real, and not unifiable — so it is now specified

Confirmed factually: `pad64` (0..=64, zero-pad, `Report data is too long`) on `DstackGuest.GetQuote`, `DstackGuest.Attest` and v1 `Attest`; `content.try_into()` — exactly 64, `invalid content length` — on `Tappd.RawQuote` and `Tappd.TdxQuote(hash_algorithm="raw")`.

But relaxing `RawQuote` to pad would answer a 32-byte request that fails today with a quote over a **different `report_data`**, silently, on a frozen surface. That is worse than the inconsistency.

So it is pinned instead: a table test over 5 surfaces × {0, 1, 63, 64, 65}, verified non-vacuous by flipping one cell. And the proto is corrected — `RawQuoteArgs.report_data` said "64 bytes of report data", which was true for one of the three methods that take it.

### v1 `Health` names containers ungated, deliberately

The only non-test reader of `HealthResponse.unhealthy` outside the guest agent is the gateway's `describe_unhealthy`; its output becomes `Observation.reason` and is printed by `record_instance_health`. Routing uses `healthy` alone — but the names *are* consumed, by the one line that tells a gateway operator which container held an instance out of rotation, with real machinery built for it (`MAX_NAMED_CONTAINERS`, the "(and N more)" tail, `sanitize`/`MAX_REASON_BYTES`, and a test named `unhealthy_containers_are_named`). Gating would leave that line permanently reading "agent reported unhealthy without naming anything".

And it cannot be gated for strangers but not the gateway: the external listener is a **single `0.0.0.0:8090` socket**, so that means putting authentication on a listener that deliberately has none.

The disclosure is also narrower than it looks — only for an app that set `requirements.health_check`, only for currently-unhealthy containers, and **not at all** when the app declared a `health_status_file`, where the path emits the constant `"<health_file>"`.

No code change; the reasoning is recorded on the field in `agent_rpc_v1.proto`.

## How this composes with the in-flight PRs

All five were test-merged. **#1232, #1254 and #1255 merge clean.** Two conflict, both mechanical, and each was resolved on a throwaway branch to prove the result:

**#1256** — two conflicts in `rpc_service.rs`. Adjacent insertions after `AppIdentity` (keep both), and `get_info`'s body: keep this structure and put their `attest_off_executor(platform, |p| p.attestation_for_info()).await?` inside the new `else` branch. Resolved tree: **154 tests pass** (153 + their `a_blocking_quote_does_not_park_the_runtime`).

Both findings survive #1256, which is worth stating because it moves the blocking off the executor but **changes no call counts**. For the demo cert, its `issue_cert` hunk moves that quote onto the blocking pool, so N unguarded spawns become up to N of tokio's 512 blocking threads plus N KMS connections instead of N parked workers — identical amplification, still needs the guard. For `Info`, after this change the hidden path has no `attestation_for_info()` left to wrap and the serving path keeps their hop exactly; their `identity()` comment ("stays on the executor; the retry throttle caps it at one quote per interval") now covers strictly more traffic, which is the direction it argues for.

**#1240** — one conflict in `process.rs`: both append a test module at EOF, keep both. Resolved tree passes 10 tests. Their `[dev-dependencies] tempfile.workspace = true` is byte-identical to this one, so `Cargo.toml` merges without conflict.

## One judgement call worth a maintainer's eye

For finding 2, apps with `public_tcbinfo = true` still attest per external `Info`. The documents genuinely are attestation data, and with #1256 that call is off the executor — but it does mean an app that published its TCB info can still have quote generation driven anonymously. Caching `tcb_info` too is possible (every field in it is boot-immutable) and would close it, at the cost of a larger frozen-surface behaviour change. Judged out of scope for this finding.


kvinwang:
Follow-up commit 5d3ebe1322 also moves the residual startup-identity retry path off the executor and coalesces concurrent retries; cancellation retains the guard.

PR #1265, opened 2026-09-20
fix(gateway): the reload does 2N point reads under the routing lock, and the health poller's restart loop cannot run

kvinwang:
Five audit findings worked to a conclusion: **four confirmed and fixed, one refuted.** The refutation lands as a corrected rationale rather than nothing.

`cargo test -p dstack-gateway`: 313 passed. `cargo clippy -p dstack-gateway -- -D warnings --allow unused_variables`: clean.

### 1. The instance reload does 2N single-key reads under the routing mutex

`read_overrides` did two `persistent.read()` point reads per instance — `instance_gate` and `instance_port_policy_override` — from the loop that runs **after** `proxy.lock()`. `start_wavekv_watch_task` re-runs the reload on every change to the `inst/` **or** `admin/` prefix from any node, so one registration anywhere costs O(fleet) store-lock acquisitions under the routing mutex on every node, and a fleet re-registering makes a refresh cycle O(fleet²). Nothing bounded it.

```
reloading 256 instances took 512 single-key override reads against 16 for 8:
the reads still scale with the fleet, under the routing lock
  left: 16   right: 512
```

New `KvStore::load_all_instance_overrides()` does one pass over `admin/`, in the same shape as the existing `legacy_instance_overrides`. It preserves the three-state read — absent / value / undecodable — through a `StoredOverrides` type, because folding "corrupt" into "nobody gated this" is the one mistake available here: a corrupt `admin/<id>/ready` must keep holding the instance out of rotation, and `a_cleared_override_is_an_answer_not_an_absence` and `one_override_key_existing_does_not_answer_for_the_other` still pass against the bulk path.

Called once *before* the lock. Reload point reads: **2N → 0**.

### 2. The health poller's restart loop was dead in the only profile that ships

`[profile.release] panic = "abort"`. The `Err(err) => error!("health poller died; restarting")` arm needs a `JoinError`, which under `abort` can never be produced; the handle was local so it could not be cancelled either; and `poll_forever` diverges, so the `Ok(())` arm was unreachable too. The whole supervision loop was dead, and its stated reason — *"a poisoned `ProxyState` mutex is enough, since locking it panics"* — was false.

**There is no failing test for this, and that is the finding.** The arm is unreachable by construction, so no test can execute it. Reporting it that way rather than dressing it up.

**Nine `or_panic` sites inherit the same reasoning**, including `ProxyInner::lock` — the routing mutex every caller funnels through — plus `proxy/port_policy.rs` ×2, `proxy.rs` ×3, `proxy/splice.rs` ×2 and `cert_store.rs`.

Replaced with a bare `tokio::spawn(poll_forever(...))` and a comment stating the mechanism and naming the real supervisor (`restart: always` in the app's compose file). Added `the_release_profile_aborts_on_panic`, which parses the workspace manifest and fails if `panic = "abort"` is ever removed — otherwise that switch would silently turn nine `or_panic`s into half-dead tasks in a process that keeps serving.

### 3. `AcmeInfo`'s bound was applied after the decode

#1253 bounds the *response*; an unauthenticated caller still paid for the whole renewal history, because `list_cert_attestations` msgpack-decoded every history key under the persistent read lock and sorted on the decoded `generated_at`.

Measured on this box, debug build: **120.6 ms for 10 000 entries against 8.1 ms for the newest 32** — the decode is ~93 % of it.

`list_recent_cert_attestations(domain, limit)` reads the sort key out of the KV key itself (`cert/{domain}/attestation/{ts}`), selects the newest `limit`, and decodes only those; `latest` falls out of the same parse. `list_cert_attestations` stays as a `limit = 0` delegate so `Admin.ListCertAttestations` is untouched.

### 4. `Admin.GetInfo` and `GetMeta` read handshakes under the routing lock

`ProxyInner::latest_handshakes` documents the rule in its own doc comment — *"a caller that only needs handshakes has no reason to hold `ProxyState` while this builds its map… on a cold cache shells out to `wg show`"* — and these two handlers did exactly that. Measured: a cold `latest()` costs ~3 ms (50 reads = 150.8 ms), all of it under the routing mutex, once per admin call, and the dashboard reaches these handlers.

```
the routing lock was taken in 4930818 of 4980304 samples across 170ms:
the handshake read is happening under it
```
(99 % of samples; <1 % after)

Reordered to match #1253's `Admin.Status` fix.

## Refuted: the DNS-TXT route grants nothing

The premise — that an arbitrary zone owner gains something by publishing `_dstack-app-address.<sni> TXT "<instance-id>:<port>"` — does not hold.

`proxy.rs:193-207` splits any inbound SNI into `subdomain.base_domain`, and when the gateway holds a wildcard for its own base domain, `parse_dst_info` hands the label straight to `select_top_n_hosts` — **the same function, the same instance-id branch, reached by any internet client with no DNS control at all.** `<instance-id>-443s.<base_domain>` is byte-for-byte the TXT route. Both paths then apply `filter_allowed_addresses` identically, so the TXT route is not more permissive in any respect.

And the behaviour is the documented contract of the very API the finding says is bypassed — `gateway/rpc/proto/gateway.proto:591-603`:

> Only affects connections addressed to the app id. Connections addressed to the instance id are never gated, so an instance pulled from rotation stays directly reachable for investigation.

Repeated in `models.rs:246-249`, `docs/app-health-checks.md:180-182`, and `CHANGELOG.md:33` (released). Consumers a gating change would break, all checked: the frozen proto contract, and `gating_an_instance_removes_it_from_rotation_but_keeps_it_directly_reachable` and `an_unhealthy_instance_is_still_reachable_by_instance_id`.

What the finding *did* correctly identify is that `models.rs`'s rationale read as if the operator were the only caller. It now states that an instance id is a routing label and not a credential, that the gate is a load-balancing control and never a way to stop an instance serving, and names what does: stopping the CVM, or a port-policy lockdown.

## Merge with the in-flight PRs

Every hunk of #1235, #1239, #1245 and #1253 was checked against this diff. **Exactly one textual conflict**, in `main_service.rs` `acme_info`, with #1253's `@@ -664,9 +721,12 @@`. Resolution after #1253 lands: keep the `list_recent_cert_attestations(domain, …)` line, drop #1253's now-redundant `.take(MAX_ACME_HIST_KEYS_PER_DOMAIN)`.

What this assumes about #1253: `MAX_ACME_HIST_KEYS_PER_DOMAIN` is defined here in `kv/mod.rs` (same name, same value 32, same rationale) and referenced fully qualified, while #1253 defines a same-named constant in `main_service.rs`. After the merge both exist in different modules, so it compiles unambiguously and #1253's test keeps working — but **a reviewer should collapse them into the `kv` one**, since the bound is now a property of the walk rather than of the response.

Item 4 assumes nothing from #1253 and is strictly stronger than its `latest()` fix: that one stops the fork, this one stops holding the lock across the read regardless of what `latest` costs. Different files, both survive.

Everything else is clear, and I confirmed rather than assumed it: `kv/mod.rs:2248` sits between #1235's hunks at 2118-2153 and 2409-2420; `admin_service.rs:127/165` and the new test module at 1452 are clear of #1235's 647-700 / 930-965 / EOF hunks and #1253's 52-56; `models.rs` and `proxy/health_check.rs` are untouched by all four.

## One clean follow-up, deliberately not done here

`Admin.ListCertAttestations` still decodes the full history and truncates afterwards. Fixing it means passing `request.limit` into the store — which is the exact line #1235 rewrites, so it is a one-line follow-up once #1235 lands rather than a conflict to create now.


kvinwang:
Follow-up commit afc2af7860 completes residual Admin.Status work: KV handshake reads and node enumeration are outside ProxyState, and status runs off the async worker.

PR #1268, opened 2026-09-20
fix(kms): a partial bootstrap is unrecoverable, a 31-byte compose hash is padded onto a whitelisted one, and the legacy backend reads unfinalized state

kvinwang:
The last eight items of the KMS audit backlog (K-a … K-h), each driven to a verdict rather than left as a note. **All eight confirmed.** Two of them resolve to something other than a code change in this diff, and both say so explicitly below rather than shipping a fix that does not hold.

## K-a — a partial bootstrap leaves the KMS unable to start *or* to re-bootstrap

Fault injection over every prefix of `Keys::store` on `next`:

```
4 writes left an unrepairable cert dir: failed to open file .../rpc-domain
a bootstrapped KMS must be able to re-read its own published record: KMS has already been bootstrapped
```

`keys_exists()` wants **7** files; the two bail conditions want **2**. Every state in between enters onboarding, where every RPC then refuses — and `Onboard.Bootstrap` refuses too, because the root keys are present. The half-root-key state reports `KMS has already been bootstrapped`. Worse, `bootstrap-info.json` was written *before* the keys, so the pubkeys registered on-chain could have had no private key behind them.

**Fix.** `config::RootKeys` becomes the one definition of "bootstrapped" (root-ca.key + root-k256.key). `Keys::store` writes an ordered `artifacts()` list — `rpc-domain` first because it is unreconstructable operator input, the two root keys adjacent, certificates last — so only **one** window is terminal instead of five, and that one names both files and the recovery. `update_certs` rebuilds a missing certificate rather than letting the KMS fall back into onboarding. `Onboard.Bootstrap` now writes keys before the record and re-serves the record for an already-bootstrapped KMS, refusing a *different* domain.

**Compat.** `Onboard.Bootstrap` becomes idempotent for the same domain, returning only public material `GetMeta` already serves. `keys_exists()` is unchanged, so #1237's `bootstrap_keys` guard and its test still hold — but #1237's comment *narrative* ("sends the KMS back here") needs a one-line update on merge.

## K-g — a 31-byte compose hash is left-padded onto a whitelisted 32-byte one

```
short_identities_are_refused_before_they_reach_the_backend
  called Result::unwrap_err() on an Ok value: BootResponse { is_allowed: true, ... }
```

A 31-byte compose hash is accepted and `padStart`ed onto the whitelisted 32-byte hash with a `0x00` first byte. Separately, `DstackApp.addComposeHash(bytes32(0))` succeeded, while `_initializeCommon` skips zero.

**Fix.** `ensure_identity_widths` at `AuthApi::is_app_allowed` — the single point every backend call goes through, so it covers the SNP path without touching #1247's hunk. `DstackApp.addComposeHash`/`addDevice` reject `bytes32(0)`.

**Compat.** The trusted OS always emits a 32-byte sha256 and a 20-byte-or-empty instance id, so no real deployment changes; `--no-instance-id` stays valid. The Solidity guard blocks only *new* zero entries.

## K-d — concurrent boots each pay their own upstream round trip

```
24 concurrent anonymous GetMeta calls cost 24 upstream calls
```

1 with the fix.

**Fix.** `InFlight<T>` coalesces requests that are identical *and* overlapping in time. **Explicitly not a cache** — the entry is retired before the result is published, so `repeated_authorization_is_never_served_from_a_decision_cache` still passes. `send_request`/`http_get`/`http_post` are byte-for-byte untouched (#1237's hunk).

**Scope, stated in the doc comment and the commit message:** this bounds concurrency amplification, **not** the serial rate. Bounding the serial rate needs a listener-level limiter the KMS does not have.

## K-h — a disabled image verification is invisible

`image.verify = false` was logged at `info!`, while the comparable `enforce_self_authorization = false` warns. `BootInfo.os_image_hash` then comes straight from the caller's `vm_config`.

**Fix.** Startup `warn!` on the same terms, the per-call line demoted to `debug!`, and `GetMeta.os_image_verification` (proto field 10, `optional bool`) so a client can tell. Additive optional field; an older KMS leaves it unset, which the proto comment says must not be read as `true`. No SDK consumes `GetMeta` today.

## K-c — the onboarding `os_image_hash` fallback is a weakening, not a break

The comment's claim holds: `mr_aggregated = sha256(mr_td‖rtmr0‖rtmr1‖rtmr2‖rtmr3)` is strictly finer than `os_image_hash`. But the substituted value is by construction one the local KMS already accepts, so `allowedOsImages` decides nothing about the source. **The operational consequence is the finding: retiring a KMS image requires removing its `mrAggregated`, not just its `allowedOsImages` entry.**

**Fix.** Warn when the fallback fires; spell out the `kmsAllowedAggregatedMrs` dependency and the revocation consequence in code; narrow the fallback to attestations carrying *no config at all* — a source that presents a config and still reports no image now keeps an empty hash and is denied, since `allowedOsImages[bytes32(0)]` is false. Legacy-format sources keep working.

## K-f — the legacy `auth-eth` backend reads unfinalized chain state, on an unpinned chain

Replayed `auth-eth-bun`'s finalized-snapshot suite against the legacy backend as `src/ethereum.test.ts`: **7 failed, 7 total.**

Found along the way, and arguably the larger finding: **`jest.config.js` matched only `main.test.ts`, so `ethereum.ts` had never been exercised at all.**

**Fix.** Ported `ETH_CHAIN_ID` pinning and `ETH_FINALITY_CONFIRMATIONS`, with the same names and parsing as the bun backend; every read in one decision now uses the same `blockTag`. Defaults (`undefined` / `0n`) preserve current behaviour.

**Note for the release:** `dstack-app/docker-compose.yaml` builds from pinned commit `78057c9`, so it needs a pin bump to pick this up. Recorded in the README.

## K-e — the external topology ships an unauthenticated plain-HTTP auth webhook

`compose-simple.yaml` wires the bare client to a backend its own header says runs *outside* the CVM, and both `deploy-simple.sh` and the auth-simple README showed `http://`.

**This warns rather than refuses, and the reason is topology:** the shipped default `kms.toml` is `url = "http://auth-api:8000"`, so a hard HTTPS requirement would break the in-CVM deployment.

**Fix.** The KMS warns at startup on a plain-HTTP webhook, exempting loopback and single-label hosts — a compose service name cannot be a public DNS name, so that exemption is exactly the two in-CVM topologies. `deploy-simple.sh`, which *knows* it is the external topology, refuses anything but `https://`. Compose header, `.env.simple` template and README updated.

**Compat — one deliberate break:** `deploy-simple.sh` now hard-fails an `http://` `AUTH_WEBHOOK_URL`. The documented path changes, and the template changes with it. The stronger fix (a KMS-pinned response signature) needs all three backends and is recorded in `security-model.md`.

## K-b — `requireTcbUpToDate` is vacuous on AWS NitroTPM; the fix is an ABI change and is *not* in this diff

```
aws_nitro_tpm_tcb_status_is_synthesized_and_no_backend_can_tell
```

`tcb_status == "UpToDate"` is synthesized; `teeVariant == "dstack-aws-nitro-tpm"` is present in the payload and **absent from every ABI tuple**. `DstackApp.isAppAllowed` compares that string, and `auth-simple` parses `teeVariant` and never reads it. So `requireTcbUpToDate = true` passes vacuously.

**What was deliberately not done.** Adding `teeVariant` to `IAppAuth.AppBootInfo` is the real fix, and it is a deployment event across every registered app — not something to slip into a residuals PR. Stripping the synthesis instead would fail NitroTPM closed on all three backends, in a crate outside the KMS, against a behaviour the verifier documents.

**What shipped:** the pinning test above, a startup `warn!` when `aws_nitro_tpm_key_release` is enabled — it is the only remaining gate, and it is the *operator's*, not the app owner's — and the reasoning in `kms.toml`, `config.rs` and `security-model.md`. No behaviour change; `aws_nitro_tpm_key_release` still defaults false.

## Verification

```
cargo test -p dstack-kms                                          → 56 passed
cargo clippy -p dstack-kms -- -D warnings --allow unused_variables → clean
cargo fmt --check                                                  → clean
forge test --ffi                                                   → 61 passed
auth-eth jest 12 · auth-eth-bun 23 · auth-simple 16 · auth-mock 12
```

## In-flight PRs

I collided with #1237 twice and both collisions are gone: `InFlight` is generic over the response type and wraps `http_get`/`http_post` from the outside, leaving `send_request` byte-identical, and my `security-model.md` sections moved above the "KMS mTLS" heading instead of appending to the same anchor. Test merges against **#1237, #1247 and #1255 are all conflict-free**.


kvinwang:
Split into independent PRs against `next`:

- K-a #1309 partial bootstrap recovery
- K-g #1310 compose_hash / instance_id width checks
- K-d #1311 in-flight auth api coalescing
- K-h #1312 os image verification visibility
- K-c #1313 onboarding os_image_hash fallback scope
- K-f #1314 auth-eth chain ID / finality pinning
- K-e #1315 plain-HTTP auth api in the external topology
- K-b #1316 NitroTPM requireTcbUpToDate docs

PR #1272, opened 2026-09-20
fix(vmm): offload boot and periodic log rotation

### LINKED ISSUE #1270: vmm: boot and periodic log rotation copy files on async workers
## Problem
App::rotate_oversized_logs calls synchronous copytruncate (fs::copy, truncation, retention cleanup) directly from the Tokio maintenance task. The VM startup path also rotates logs synchronously. Large logs, many VMs, or slow storage can occupy async workers and delay unrelated control-plane work.

The default periodic threshold is 4 MB with a 5-second scan interval; files can grow beyond the threshold between scans. Actual latency depends on storage and workload and has not been benchmarked.

## Fix
Offload both rotation paths to blocking workers. Keep rotations serialized with each other and VM lifecycle operations, retaining that protection if a caller is cancelled. Preserve copytruncate and retention behavior and add regression tests.

#1263 changes supervisor log flushing; its VMM change is only a comment and does not offload rotation.

kvinwang:
## Problem
Boot-time and periodic VMM log rotation synchronously copy/truncate files on async workers. Slow storage or oversized logs can delay unrelated control-plane work. #1263 fixes supervisor flushing, not these rotation paths.

## Fix
- Run both rotation paths on blocking workers.
- Keep the per-VM lifecycle guard with the worker, including after cancellation; return it to the startup path afterward.
- Periodic scans skip busy/removing/unloaded VMs instead of waiting behind a long launch or teardown.
- Preserve copytruncate, retention, and legacy serial-log eligibility.

Fixes #1270. Independent of #1271; both reuse the existing lifecycle lock.

## Validation
Linux x86_64, Rust 1.92.0, debug tests; no QEMU/TDX VM or storage-throughput benchmark:
- cargo test -p dstack-vmm: **212 passed, 3 existing root/network-namespace tests ignored**.
- cargo fmt -p dstack-vmm -- --check and CI-style production clippy passed.
- A single-thread Tokio test stalls rotation on a channel (5-second failure guard), checks executor responsiveness, cancels the caller, and verifies lifecycle exclusion.
- Temporary-file tests cover periodic rotation, busy/removing/deleted VMs and legacy serial logs; existing copytruncate/retention tests also pass.

This removes rotation's blocking work from async workers; it does not accelerate the underlying copy or claim a production speedup percentage.


kvinwang:
Closing: logs rotate at 4 MiB by default so the copy is milliseconds; the lifecycle lock hand-offs and skipped rotations this adds cost more than they save.

Issue #1259, opened 2026-09-20
kms: performance findings — eight serialized round trips on every CVM boot


kvinwang:
A read of the KMS Rust service, all four auth backends, both contracts, and the verifier/`ra-rpc` paths the KMS drives, looking for cost rather than correctness. No benchmarks were run — every number below is a complexity class, a counted round trip, or an estimate labelled as such.

Readability and maintainability findings from the same sweep are collected separately; this issue is only the parts that cost time or bandwidth.

Two findings from the sweep turned out to be correctness bugs already fixed in #1247 (the `auth-simple` empty-device-allowlist fail-open, and the SNP `device_id` 64-vs-32-byte mismatch), so they are not repeated here.

## The headline: ~8 serialized network round trips per `GetAppKey`

`main_service.rs:360-369`, `auth-eth-bun/index.ts:131-198`

`get_app_key` makes **two** sequential POSTs to the auth API — `ensure_self_allowed()` → `POST /bootAuth/kms`, then `ensure_app_boot_allowed()` → `POST /bootAuth/app`. And `EthereumBackend.checkBoot` makes **four** sequential chain RPCs each:

```ts
const chainId = await this.client.getChainId();      // 1
const head    = await this.client.getBlockNumber();  // 2
response      = await this.client.readContract({ isAppAllowed, blockNumber }); // 3
gatewayAppId  = await this.client.readContract({ gatewayAppId, blockNumber }); // 4
```

That is **8 strictly serialized Ethereum RPC round trips on every CVM boot**, each preceded by a fresh TCP/TLS connect (see below). Against a hosted endpoint at a plausible 80–150 ms RTT that is 0.6–1.2 s of boot latency, all of it avoidable.

Three of the four are unnecessary as written:

- `getChainId()` is invariant for a given endpoint — a startup assertion, not a per-decision one.
- The two `readContract`s are pinned to the **same** `blockNumber` and are independent: `await` in a loop where `Promise.all` is correct.
- viem's HTTP transport supports `http(rpcUrl, { batch: true })`, which coalesces JSON-RPC calls issued in the same tick into one POST — turning steps 3+4 into a single request.

*Fix:* assert `chainId` once at startup, `Promise.all` the two reads, construct the transport with `{ batch: true }`. 4 sequential round trips → 2. ~15 lines.

[...]

PR #1277, opened 2026-09-20
fix(verifier): offload and coalesce cold image verification

kvinwang:
Completes the cold verifier findings in #1259/#1261. Coalesces same-image downloads and measurement work, offloads extraction/checksum/measurement, bounds blocking work, and protects publication from cancellation/races. Deterministic test: baseline 16 concurrent downloads, fixed 1.


kvinwang:
Closing: the measurement offload landed via #1336 and download coalescing is covered more simply by #1337; the remaining global 2-slot semaphore would queue every verify request, including cheap ones.

PR #1306, opened 2026-09-22
fix(kms): pin onboarding key fetch to the authorized source KMS

kvinwang:
Split from #1237.

Onboarding verifies the source KMS on one TLS connection, then fetches root keys over a second `new_mtls` connection that does not authenticate the server. Pin the second connection to the SPKI from the attested certificate authorized on the first connection.

This prevents an endpoint substitution from poisoning a new KMS with unrelated root keys. A correctly configured relying party would reject those unrelated keys against the trusted KMS identity, so the practical impact is onboarding integrity and availability rather than bypass of the final trust chain.

The now-unused `RaClient::new_mtls` helper is removed. A regression test verifies that the key-fetch connection rejects a different server.

Verification: `cargo fmt --check`; `cargo test -p dstack-kms`.


kvinwang:
Closing for now. #1108 provides the cleaner fix by using a single attestation-verified connection for the key fetch. Also, endpoint substitution here does not bypass the final relying-party trust check: unrelated root keys do not match the trusted KMS identity, so the remaining impact is onboarding integrity/availability rather than a core trust-chain compromise.

PR #1315, opened 2026-09-22
fix(kms): refuse an unauthenticated auth api in the external topology

kvinwang:
The KMS reaches the auth webhook with a bare HTTP client, and that answer is the key-release decision. `compose-simple.yaml` wires it to a backend running outside the CVM, and `deploy-simple.sh` and the auth-simple README showed `http://`.

- The KMS warns at startup on a plain-HTTP webhook, exempting loopback and single-label hosts (the in-CVM topologies, including the shipped `http://auth-api:8000`).
- `deploy-simple.sh`, which is always the external topology, refuses anything but `https://`. Compose header, env template and README updated.

Breaking: `deploy-simple.sh` now fails on an `http://` `AUTH_WEBHOOK_URL`. A KMS-pinned response signature would be the stronger fix; recorded in `security-model.md`.

Split out of #1268 (K-e).


kvinwang:
Closing. The strictest setup runs the auth backend (e.g. a light client) as another container in the same compose, which is plain `http://` by design, so a URL-scheme heuristic adds noise without adding a guarantee. The webhook backend is operator-controlled in any case; a real fix would be an attested or signed response, not a scheme check.

